PR #36, opened 2026-04-03
fix: skip numpy-dependent MLX tests when numpy not installed

maziyarpanahi:
## Problem
macOS CI failed with `ModuleNotFoundError: No module named 'numpy'` in 3 MLX conversion tests. The `[dev]` extras don't include numpy, and these tests ran on a minimal CI environment.

## Fix
- **Moved** `test_remap_all_hf_keys` from `TestConvertEndToEnd` to `TestWeightKeyRemapping` — it only tests string remapping and never needed the numpy/torch fixture
- **Added** `pytest.importorskip("numpy")` before `TestSaveNumpyModel` — these tests are cleanly skipped when numpy isn't installed
- **Removed** the `mock_hf_model` fixture (was the source of the `torch` and `numpy` imports that caused the setup error)
- **Added** `test_creates_parent_dirs` test for better coverage

## Test plan
- [x] `pytest tests/unit/mlx/test_mlx_convert.py -v` — 18 passed locally (numpy available)
- [x] `pytest tests/ -q` — 700 passed, 0 failures
- On CI without numpy: 15 key-remapping tests pass, 3 numpy tests skip gracefully

### DIFF
--- a/tests/unit/mlx/test_mlx_convert.py
+++ b/tests/unit/mlx/test_mlx_convert.py
@@ -6,6 +6,7 @@ remapping and config extraction logic in isolation.
 
 from __future__ import annotations
 
+import json
 import pytest
 
 from openmed.mlx.convert import remap_key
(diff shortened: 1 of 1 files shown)


maziyarpanahi:
Superseded by #39, which consolidates the 1.0.0 public release work into one PR.

PR #49, opened 2026-05-03
Add privacy filter stream demo

maziyarpanahi:
## Summary

- Add a FastAPI example that streams raw agent traces immediately while emitting a delayed share-safe stream.
- Include deterministic masking or Faker-backed replacement for sensitive trace values such as names, contact details, IDs, account numbers, network indicators, and secrets.
- Add the static frontend, logo, run instructions, and docs index entry for the new examples/privacy_filter_stream/ demo.

## Why

This gives developers a concrete, browser-based example of how OpenMed privacy filtering can sit between an internal agent trace and a shareable trace export. The side-by-side stream makes the delay, redaction mode, and repeated-value consistency easy to inspect without downloading a model.

## Impact

The new example is model-free and cache-free, so it starts quickly for docs and demos. It does not change library behavior outside the example directory and docs listing.

## Validation

- Passed: Python AST parse for app.py and __init__.py with python3.
- Passed: node --check examples/privacy_filter_stream/static/app.js.


maziyarpanahi:
Closing this stale implementation during the post-v2.0 reconciliation. It is more than 2,000 commits behind current master and no longer merges cleanly. No verified roadmap issue is linked, so no issue state is being changed.

PR #57, opened 2026-05-25
fix: cast bfloat16 tensors to float32 before numpy conversion in MLX pipeline

kswanjitsu:
## Problem

When loading models with bfloat16 weights (e.g. `OpenMed/privacy-filter-nemotron`) through the MLX pipeline, conversion fails with:

```
TypeError: Got unsupported ScalarType BFloat16
```

This happens in `_to_numpy()` because PyTorch CPU does not support `.numpy()` on bfloat16 tensors.

## Fix

Cast bfloat16 tensors to float32 before calling `.numpy()`. The resulting MLX weights are stored in float32, which is fully supported on all Apple Silicon chips and incurs no meaningful accuracy loss for token classification.

## Affected models

- `OpenMed/privacy-filter-nemotron` (and any future bf16 models)

## Testing

Verified that `create_mlx_pipeline("OpenMed/privacy-filter-nemotron")` completes weight conversion successfully after this fix.

PR #58, opened 2026-05-25
fix: add MLX weight remapping for openai_privacy_filter / nemotron architecture

kswanjitsu:
## Problem

The `openai_privacy_filter` model family (e.g. `OpenMed/privacy-filter-nemotron`) cannot be loaded via `create_mlx_pipeline` because its HuggingFace weight namespace does not match the MLX model's namespace. All 140 parameters are rejected with `ValueError: Received 140 parameters not in model`.

## Root cause

The existing `remap_key()` function handles BERT/RoBERTa/DeBERTa/DistilBERT/ELECTRA architectures but has no case for `openai_privacy_filter`. The OPF architecture differs from all others:
- Different top-level prefix (`score.*` → `unembedding.*`)
- Different layer container (`model.layers.N.*` → `block.N.*`)
- QKV fusion: HF has separate `q/k/v_proj`, MLX has single `attn.qkv`
- Different sub-key names (`input_layernorm` → `attn.norm`, `router` → `gate`)
- RMSNorm uses `scale` not `weight`
- Expert weights already in `[E, in, out]` format (no transpose needed)

## Fix

Add `_convert_opf_weights()` and wire into `convert_weights()` for model types `openai-privacy-filter`, `privacy-filter-nemotron`, `nemotron-privacy-filter`. Also set `classifier_bias: True` in the MLX config when `score.bias` is present.

## Testing

Verified on `OpenMed/privacy-filter-nemotron` — model now loads and runs inference correctly:
```python
pipe = create_mlx_pipeline("OpenMed/privacy-filter-nemotron")
result = pipe("Patient John Smith DOB 01/01/1980")
# Returns: first_name=John, last_name=Smith, date=01/01/1980
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/openmed/mlx/convert.py
+++ b/openmed/mlx/convert.py
@@ -93,6 +93,16 @@ _ELECTRA_KEY_REPLACEMENTS: list[Tuple[str, str]] = [
     ("classifier.", "classifier."),
 ]
 
+_OPF_MODEL_TYPES = {
+    "openai-privacy-filter",
+    "privacy-filter",
+    "privacy-filter-nemotron",
+    "nemotron-privacy-filter",
+    "privacy-filter-multilingual",
+    "multilingual-privacy-filter",
+}
+
+
 def _convert_opf_weights(state_dict: dict) -> dict:
     """Remap and reshape HuggingFace OPF weights to the MLX namespace.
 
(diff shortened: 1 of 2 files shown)


maziyarpanahi:
Thanks again for chasing this down. I pushed one follow-up commit onto this PR so it now carries the PR #57 fix plus the OPF/Nemotron remapping in one place.

What changed in the follow-up:
- BF16 tensors are now cast to explicit float32 before `.numpy()`. This is still needed because PyTorch CPU cannot expose `torch.bfloat16` tensors through NumPy. I used `t.float()` rather than `t.to(float)`, because Python `float` promotes to float64 in PyTorch and would unnecessarily increase memory/artifact size.
- The OPF/privacy-filter converter now validates the converted MLX key set and tensor shapes before returning weights. That means a partial mapping, for example missing one Q/K/V tensor, fails early instead of producing a broken artifact.
- Added focused regression tests for BF16 conversion, OPF HF→MLX remapping, QKV fusion order, and partial-QKV rejection.

The reason this failed for local conversion but not for the existing OpenMed demo is that the demo uses the already-exported `OpenMed/privacy-filter-nemotron-mlx` artifact. That artifact is already in the OpenMed MLX runtime layout. The broken path was only raw HF → MLX conversion from `OpenMed/privacy-filter-nemotron`, where the source checkpoint uses `model.layers.*`, `score.*`, separate `q/k/v_proj`, RMSNorm `weight`, and MoE expert tensors.

I also ran the full suite locally on the updated PR branch: `1153 passed, 1 skipped`.

maziyarpanahi (approved):

PR #199, opened 2026-06-11
feat: accept all 12 PII languages in the REST service

Technolity:
What

The library ships per-language PII models for 12 languages (DEFAULT_PII_MODELS),
but the REST /pii/extract and /pii/deidentify lang field only allowed 9, so
requests for ar, ja, and tr were rejected with a 422 despite the models existing.

Define the language list once as PIILang, reuse it across all request schemas
(pydantic v1 and v2), add ar/ja/tr, and add a parity test that locks the REST
options to DEFAULT_PII_MODELS so the two cannot drift again.

Verified

Full suite: 1221 passed, 22 skipped.


maziyarpanahi:
Correction to the note above: this is closed as superseded by #210, which already merged the REST/MCP PII language parity fix against the current schema name. The current merge result fails local CI with `.venv/bin/python -m pytest tests/ -q`: `tests/unit/service/test_api.py::test_pii_lang_options_match_core_models` imports `PIILang`, while the merged schema exposes `PIILanguage`.

PR #202, opened 2026-06-11
chore: run the Docker container as a non-root user

Technolity:
What

The image ran uvicorn as root. Add an unprivileged appuser (created after the
root-only install step), give it ownership of /app and a home for the runtime
Hugging Face cache, and switch to it before EXPOSE/CMD so the service no longer
runs as root.

Verified

Confirmed the Hugging Face cache and the openmed cache_dir both resolve under
the appuser HOME, so they remain writable at runtime.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,13 @@ RUN python -m pip install --upgrade pip \
     && pip install --no-cache-dir --index-url https://download.pytorch.org/whl/cpu torch \
     && pip install --no-cache-dir ".[hf,service]"
 
+# Run as an unprivileged user. Created after install (which needs root); the
+# --create-home flag gives appuser a writable HOME for the Hugging Face model
+# cache pulled at runtime.
+RUN useradd --create-home --uid 1000 appuser \
+    && chown -R appuser:appuser /app
+USER appuser
+
 EXPOSE 8080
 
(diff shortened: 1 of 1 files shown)


maziyarpanahi:
I pushed a small follow-up that keeps the non-root change and makes runtime cache paths explicit: `HOME=/home/appuser`, `XDG_CACHE_HOME=/home/appuser/.cache`, and `HF_HOME=/home/appuser/.cache/huggingface`, with the OpenMed and Hugging Face cache directories created and owned by `appuser`.

Current merge-result local CI passes with `.venv/bin/python -m pytest tests/ -q`: `1286 passed, 1 skipped`.

I am leaving this open for now rather than merging because the Docker-specific smoke check could not be run locally: the Docker CLI is installed, but both available Docker contexts point at a non-running daemon. This should be merged after a Docker build/run check confirms the image starts as non-root and can write to the OpenMed and Hugging Face cache paths.

maziyarpanahi:
Closing this stale PR so the active queue reflects work that can still be completed. The non-root container change remains unlinked to a maintainer-backed issue, and the Docker runtime validation requested in the prior review was never completed against the current image. Thank you for the contribution and the cache-path follow-up. A fresh issue-scoped PR against current master can revisit it.

PR #208, opened 2026-06-12
Fix case-sensitive trust_remote_code allowlist matching

### LINKED ISSUE #205: trust_remote_code allowlist is case-sensitive; confirm intended matching
Summary

is_trusted_for_remote_code() in openmed/torch/privacy_filter.py compares the
model name against TRUSTED_REMOTE_CODE_MODELS with exact, case-sensitive
membership. A user passing a different case (e.g.
"openmed/privacy-filter-multilingual" vs "OpenMed/...") will not match, so
trust_remote_code stays False and the first-party model fails to load its
custom code.

Security caveat (why I did not just make it case-insensitive)

This allowlist gates trust_remote_code=True, i.e. arbitrary code execution from
the repo (see the CVE-2026-47117 regression tests). Whether case-insensitive
matching is safe depends on HuggingFace namespace semantics:

- if HF namespaces are case-insensitive (OpenMed and openmed are the same org),
  normalizing both sides is safe and correct
- if they can be distinct orgs, case-insensitive matching could let a lookalike
  repo be trusted

I did not want to loosen a security boundary on a guess.

Question

What is the intended behavior? If HF namespaces are case-insensitive I can send
a PR normalizing case on both sides (including the env-var allowlist) with
tests; otherwise this may be case-sensitive by design and worth a doc note.

sawirricardo:
## Summary

Fixes #205

`is_trusted_for_remote_code()` compared model names against the hardcoded allowlist with exact case-sensitive membership. Since HuggingFace namespaces are case-insensitive (`OpenMed` and `openmed` resolve to the same org), a user passing `openmed/privacy-filter-multilingual` would not match the hardcoded `OpenMed/privacy-filter-multilingual`, so `trust_remote_code` stayed `False` and the first-party model failed to load its custom code.

## Changes

- Store `TRUSTED_REMOTE_CODE_MODELS` entries in lowercase
- Normalize `model_name` to lowercase in `is_trusted_for_remote_code()` before comparison
- Normalize env-var entries to lowercase in `_env_allowlist()`
- Add tests for case-variant matches across all three layers:
  - Hardcoded allowlist (e.g. `openmed/privacy-filter-multilingual`, `OPENMED/PRIVACY-FILTER-NEMOTRON`, `OpenAI/Privacy-Filter`)
  - Env-var allowlist (`My-Org/My-Fork` matches `my-org/my-fork`)
  - Pipeline gate (lowercase variant passes `trust_remote_code=True`)

## Security

This is safe because HuggingFace namespaces are case-insensitive — `OpenMed` and `openmed` resolve to the same org on the Hub. Normalizing both sides to lowercase does not expand the trust boundary to include lookalike repos from different orgs.

## Testing

All 44 tests in `test_privacy_filter_security.py` pass, including the 3 new case-insensitivity tests.

### DIFF
--- a/openmed/torch/privacy_filter.py
+++ b/openmed/torch/privacy_filter.py
@@ -25,10 +25,14 @@ logger = logging.getLogger(__name__)
 # First-party privacy-filter repos that legitimately require
 # trust_remote_code=True (they ship modeling_openai_privacy_filter.py and
 # friends in the repo and rely on Transformers' auto_map import).
+# Stored in lower-case because HuggingFace namespaces are case-insensitive
+# (OpenMed and openmed resolve to the same org).  Normalising both sides
+# prevents a user-supplied "openmed/privacy-filter-multilingual" from failing
+# to match the hard-coded "OpenMed/..." entry.
 TRUSTED_REMOTE_CODE_MODELS = frozenset({
     "openai/privacy-filter",
-    "OpenMed/privacy-filter-multilingual",
-    "OpenMed/privacy-filter-nemotron",
+    "openmed/privacy-filter-multilingual",
+    "openmed/privacy-filter-nemotron",
 })
 
 # Operators with custom fine-tunes can extend the allowlist with a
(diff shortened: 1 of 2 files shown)


maziyarpanahi:
Thanks @sawirricardo for the PR, and thanks @Technolity for raising the edge case.

I verified the issue against the implementation, checked that case variants resolve to the same first-party repository on the model hub, and ran the required local tests:

- `.venv/bin/python -m pytest tests/unit/test_privacy_filter_security.py -q` — 44 passed
- `.venv/bin/python -m pytest tests/ -q` — 1265 passed, 1 skipped

No extra code changes were needed; the branch is ready to merge. This fix will be included in the next release.

maziyarpanahi (approved):

PR #371, opened 2026-06-18
fix: use unique placeholders for repeated entity types in deidentify() (closes #222)

### LINKED ISSUE #222: Fix reidentify() round-trip when an entity type repeats
## Summary
reidentify() cannot round-trip when the same entity type appears more than once (addresses #204). In core/pii.py the mapping built during de-identification is keyed by the redacted placeholder string (line 729: mapping[redacted] = entity.original_text or entity.text). With method='mask' and several entities sharing a label (e.g. two PERSONs -> two '[NAME]' placeholders, or two DATEs), the second write clobbers the first, so the dict holds only one original. reidentify() (line 1446) then does a global str.replace(redacted, original) for each key, restoring EVERY occurrence of that placeholder to the single surviving original. Result: deidentify(text, keep_mapping=True) followed by reidentify() does not reconstruct the source (the docstring example 'assert original == text' fails for repeated types). The data needed to round-trip (per-span original_text + start/end) is already on each PIIEntity; the placeholder-keyed dict is the lossy step.

## Scope
- [ ] Make the keep_mapping path lossless for repeated entity types: either index placeholders so each occurrence is uniquely keyed (e.g. append an occurrence counter such as '[NAME_1]'/'[NAME_2]' in the produced text and mapping) OR change the mapping to a positional/ordered structure (list of {placeholder, original, start, end}) that reidentify() consumes left-to-right by offset rather than global string replace.
- [ ] Update reidentify() in core/pii.py to consume the new structure deterministically and reconstruct the exact original for repeated placeholders; keep the existing (placeholder->original) dict signature working for back-compat where the placeholders are already unique, or document and version the mapping shape.
- [ ] Ensure consistency across methods that emit identical placeholders for distinct originals (mask, remove, hash, replace) and that surrogate ('replace') re-identification still maps each surrogate back to its own original.
- [ ] Add regression tests covering: two PERSONs with different names, two DATEs, mixed repeated + unique types, and a mask/replace/hash round-trip each asserting reidentify(deidentify(text).deidentified_text, mapping) == text.

## Acceptance criteria
- [ ] deidentify(text, method='mask', keep_mapping=True) on text containing two distinct entities of the same label, then reidentify(result.deidentified_text, result.mapping), reproduces the original text exactly.
- [ ] The docstring example round-trip (assert original == text) holds for inputs with repeated entity types under mask, hash, and replace methods.
- [ ] A regression test for #204 (two PERSONs / two DATEs) fails on the pre-fix code and passes after the fix.
- [ ] Existing pii/reidentify tests remain green (no behavior change for already-unique placeholders).

## Out of scope
- Audit-report provenance fields (OM-011).
- The deterministic safety sweep (OM-008).
- Changing default placeholder text/formatting beyond what is needed to disambiguate repeats.

## Files
- openmed/core/pii.py
- tests/unit/core/test_pii_reidentify.py

---
Task: OM-057  ·  Milestone: v1.7  ·  Priority: P1  ·  Size: M
Depends on: —  ·  Blocks: —
Roadmap: open issue #204; section 5.10 reversibility/round-trip correctness
Spec: PLANS/V2/EXECUTION/tasks/OM-057.md

rtmalikian:
## Summary
Fixes `reidentify()` round-trip behavior for repeated entity types when `deidentify(..., keep_mapping=True)` stores reversible mappings.

The branch keeps plain mask/remove output unchanged when `keep_mapping=False`, assigns repeated placeholders in document order for reversible mappings, and includes regression coverage for mask, remove, hash, and replace round-trips.

Closes #222.

## Verification
- `.venv/bin/python -m pytest tests/unit/test_pii.py -q` (112 passed)
- `.venv/bin/python -m pytest tests/ -q` (1557 passed, 1 skipped)

### DIFF
--- a/openmed/core/pii.py
+++ b/openmed/core/pii.py
@@ -1070,6 +1070,7 @@ def _build_deidentification_result(
 
     deidentified = text
     mapping = {} if keep_mapping else None
+    entity_type_counts: dict[str, int] = {}
 
     for entity in redaction_entities:
         entity_method = _entity_redaction_method(entity, effective_method)
@@ -1083,6 +1084,23 @@ def _build_deidentification_result(
             lang=lang,
             anonymizer=anonymizer,
         )
+
+        # When the same placeholder would be produced for multiple entities
+        # (e.g. two PERSONs both masking to "[PERSON]"), append an occurrence
+        # counter so each mapping key is unique and round-trip is lossless.
+        if entity_method in ("mask", "remove") or (
+            entity_method == "shift_dates" and entity.entity_type != "DATE"
+        ):
+            etype = entity.entity_type
+            entity_type_counts[etype] = entity_type_counts.get(etype, 0) + 1
+            if entity_type_counts[etype] > 1:
+                # Insert counter before closing bracket for mask (e.g. [NAME_2])
+                # or append for non-bracket placeholders (e.g. remove returns "")
+                if redacted.endswith("]"):
+                    redacted = f"{redacted[:-1]}_{entity_type_counts[etype]}]"
+                elif redacted:
+                    redacted = f"{redacted}_{entity_type_counts[etype]}"
+
         entity.redacted_text = redacted
         if reversible_ids:
             entity.reversible_id = _build_reversible_id(
--- a/openmed/core/pii.py
+++ b/openmed/core/pii.py
@@ -1070,7 +1070,20 @@ def _build_deidentification_result(
 
     deidentified = text
     mapping = {} if keep_mapping else None
-    entity_type_counts: dict[str, int] = {}
+    entity_occurrence_indexes: dict[int, int] = {}
+    if keep_mapping:
+        entity_type_counts: dict[str, int] = {}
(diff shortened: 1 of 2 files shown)


maziyarpanahi:
Thank you for the contribution, and welcome as a first-time contributor here. I reviewed this against #222 and the OM-057 goal, then pushed a small maintainer follow-up to the same branch while preserving your fix.

What changed in the follow-up:
- kept the unique-placeholder approach for reversible mappings
- limited placeholder numbering to `keep_mapping=True`, so ordinary mask output remains unchanged
- assigned repeated placeholders in document order
- added regression coverage for replace-mode round-trip and repeated mask output without a mapping
- cleaned up PR metadata and applied the issue labels

Local verification passed:
- `.venv/bin/python -m pytest tests/unit/test_pii.py -q` -> 111 passed
- `.venv/bin/python -m pytest tests/ -q` -> 1556 passed, 1 skipped

GitHub is not reporting a check rollup for the fork branch yet after the maintainer push, but the branch is mergeable locally and against the PR base.

maziyarpanahi (approved):

PR #393, opened 2026-06-19
fix: consolidate validation, HTML, and release gate fixes

### LINKED ISSUE #391: fix: _contains_suspicious_content rejects valid Japanese/CJK clinical text
## Problem

The _contains_suspicious_content function in openmed/utils/validation.py uses the regex pattern to detect binary content. However, this also matches legitimate CJK clinical text which does not use spaces between words.

Japanese (ja) is an explicitly supported PII language yet any Japanese clinical note longer than about 50 characters will be rejected as suspicious content.

The same issue affects Chinese and any CJK text without spaces.

## Proposed Fix

Replace the blanket non-ASCII rejection with a check that excludes CJK ranges. Alternatively remove the non-ASCII check entirely since the other two checks already catch real binary content.

## Affected Files

- openmed/utils/validation.py - _contains_suspicious_content function
- tests/unit/utils/test_validation.py - needs test for CJK text acceptance

### LINKED ISSUE #392: fix: XSS vulnerability in OutputFormatter.to_html — no HTML escaping
## Problem

The to_html method in openmed/processing/outputs.py inserts user-controlled text directly into HTML without any escaping. This creates a cross-site scripting (XSS) vulnerability when HTML output is displayed in a browser.

Specifically:
1. result.text is not HTML-escaped before being used in highlighted_text
2. entity.label is not escaped in the span class attribute and title attribute
3. entity.text is not escaped in the entity summary list
4. result.model_name is not escaped

No html.escape() or equivalent is used anywhere in the file.

## Impact

If the HTML output is rendered in a web UI (e.g., the REST service, a dashboard, or an exported report), an attacker could inject malicious JavaScript via crafted clinical text or entity labels.

This is particularly concerning for a privacy/security tool — ironic XSS in a de-identification pipeline.

## Proposed Fix

Import html.escape and apply it to all user-controlled values before embedding them in HTML:
- Escape result.text before highlighting
- Escape entity.label, entity.text, entity.confidence in the summary
- Escape result.model_name, result.timestamp

## Affected Files

- openmed/processing/outputs.py - to_html method (line ~425)

### LINKED ISSUE #396: Release gate evaluation failed
## Summary

The release gate job failed before producing a gate report.

## Failure

- `release gate evaluation failed before a report was produced: [Errno 2] No such file or directory: 'artifacts/release-candidate.json'`

rtmalikian:
## Summary
- allow CJK clinical text while still rejecting binary-like control character runs
- escape HTML formatter output without shifting entity offsets
- skip scheduled release gates when no release candidate report exists

## Verification
- .venv/bin/python -m pytest tests/ -q
- .venv/bin/python scripts/release/check_license_policy.py

Closes #391
Closes #392
Closes #396

### DIFF
--- a/.github/workflows/release-gates.yml
+++ b/.github/workflows/release-gates.yml
@@ -33,7 +33,18 @@ jobs:
           python -m pip install --upgrade pip
           pip install -e .
 
+      - name: Check for candidate report
+        id: check-candidate
+        run: |
+          if [ -f "$CANDIDATE_REPORT" ]; then
+            echo "exists=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "exists=false" >> "$GITHUB_OUTPUT"
+            echo "No candidate report found at $CANDIDATE_REPORT; skipping release gates."
+          fi
+
       - name: Run release gates
+        if: steps.check-candidate.outputs.exists == 'true'
         id: gates
         continue-on-error: true
         env:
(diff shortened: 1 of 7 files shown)


maziyarpanahi:
Closing in favor of #402, which carries the same compacted fix set on a new repository branch with a single commit.

PR #394, opened 2026-06-19
fix: escape user-controlled text in to_html to prevent XSS (Fixes #392)

### LINKED ISSUE #392: fix: XSS vulnerability in OutputFormatter.to_html — no HTML escaping

rtmalikian:
Fixes #392

## Summary

- Escapes model, timestamp, source text, entity labels, and entity summary text before emitting HTML.
- Builds highlighted HTML from escaped text segments so original entity offsets stay valid when earlier text contains escapable characters.
- Adds regression coverage for script/image/event-attribute payloads and offset preservation.

## Verification

- `.venv/bin/python -m pytest tests/unit/test_processing.py -q`
- `.venv/bin/python -m pytest tests/ -q` against a temporary merge with updated `origin/master`

### DIFF
--- a/openmed/processing/outputs.py
+++ b/openmed/processing/outputs.py
@@ -442,34 +442,35 @@ class OutputFormatter:
 
         html += f'<div class="text-content">\n'
 
-        # Highlight entities in text
-        highlighted_text = html_mod.escape(result.text)
-        offset = 0
-
         # Sort entities by start position
         sorted_entities = sorted(
             [e for e in result.entities if e.start is not None and e.end is not None],
             key=lambda x: x.start
         )
 
+        highlighted_parts: List[str] = []
+        cursor = 0
         for entity in sorted_entities:
-            start = entity.start + offset
-            end = entity.end + offset
+            start = entity.start
+            end = entity.end
+            if start < cursor or end <= start or end > len(result.text):
+                continue
 
             color = self._get_entity_color(entity.label)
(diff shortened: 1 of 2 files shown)


maziyarpanahi:
Closing in favor of #393, which now includes the HTML escaping fix together with the validation and release-gate fixes in one consolidated PR.

PR #604, opened 2026-06-23
fix: never auto-select a zero-day date shift offset

otmanm:
# Pull Request

## Description
`shift_dates` auto-selects a day offset via `random.randint(-365, 365)` when the
caller doesn't supply `date_shift_days`. That range is inclusive of 0, so roughly
1 run in 731 silently shifts every date in the document by zero days, leaving
all clinical dates (including DOBs and admission dates) completely unredacted
with no error or warning. This adds a small helper that guarantees the
auto-selected offset is non-zero, over the same ±365-day range.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition/improvement

## Changes Made
- Add `_random_nonzero_shift()` in `openmed/core/pii.py`, returning a random
  non-zero day offset in `[-365, 365]` (sign + magnitude, magnitude drawn from
  `[1, 365]`).
- Use it in place of `random.randint(-365, 365)` for the auto-selected
  `shift_dates` offset.
- Add a property test (`TestRandomNonzeroShift`, 10k + 2k draws) confirming the
  helper never returns 0 and stays within range, including a custom `high`.
- Add a regression test (`test_deidentify_auto_shift_dates_never_echoes_original_date`)
  that runs the full auto-selected `shift_dates` path 200 times with
  `keep_year=False` and asserts the original date never reappears verbatim.
  `keep_year=False` makes the assertion deterministic: shifting a calendar
  date by any non-zero number of days always changes it, so this cannot flake
  on a leap-year edge case the way a `keep_year=True` check could.

Explicit caller-supplied `date_shift_days=0` is intentionally left unchanged by
this PR; that's a deliberate request, not the accidental no-op this fixes.

## Testing
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this change with different models/inputs

Ran:
- `pytest tests/unit/test_pii.py -q` → 119 passed
- `pytest tests/ -q` (full suite) → 1701 passed, 22 skipped, 0 failed

## Documentation
- [ ] I have updated the documentation accordingly
- [x] I have added docstrings to new functions/classes
- [ ] I have updated the CHANGELOG.md

No user-facing behavior or API changes; the auto-selected offset still ranges
[...]

PR #605, opened 2026-06-24
fix: preserve slash/dash date format for 2-digit years

otmanm:
# Pull Request

## Description
`_format_date_like_original` re-detects a parsed date's original string shape
(separator + field order) so the shifted date comes back formatted the same
way it went in. Its slash- and dash-separated regexes hardcode a 4-digit
year, so a 2-digit-year input like `03/15/22` matches neither pattern and
falls through to the function's ISO default — the date comes back as
`2022-04-14` instead of something in the original `MM/DD/YY`-family shape.
Not just the year width changes: the separator and field order silently
change too.

This is also an existing inconsistency within the same function: the
dot-separated regex two lines above already accepts 2-4 digit years, so
dotted 2-digit-year dates (`15.03.22`) are already handled correctly. This PR
brings the slash and dash regexes in line with the convention the dot regex
already establishes.

## Type of Change
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test addition/improvement

## Changes Made
- Widen the slash-separated date regex in `_format_date_like_original` from
  `\d{1,2}/\d{1,2}/\d{4}` to `\d{1,2}/\d{1,2}/\d{2,4}`.
- Widen the dash-separated date regex from `\d{1,2}-\d{1,2}-\d{4}` to
  `\d{1,2}-\d{1,2}-\d{2,4}`.
- Add `TestFormatDateLikeOriginal` in `tests/unit/test_pii.py`, calling
  `_format_date_like_original` directly (see note below on why) for:
  a 2-digit-year US slash date, a 2-digit-year EU dash date, and a regression
  guard confirming 4-digit-year dates are unaffected by the widened regex.

The rendered year is still always 4 digits (`%Y` in every `strftime` call);
only the *input* detection regex changed. This matches what the sibling
`_shift_date_basic` function already does for 2-digit years, so this PR
doesn't introduce a new digit-width convention, just fixes a shape/order
regression unique to this function.

## Note: this code path has no CI coverage today
While tracing this, I noticed `_format_date_like_original` is only reachable
through `_shift_date`, which is only used when `python-dateutil` is
importable — otherwise `_shift_date` falls back to `_shift_date_basic`.
`python-dateutil` isn't in `pyproject.toml`'s `dev` extra or core
dependencies (it's only a transitive dep of `pandas`/`ghp-import`), and CI's
test job installs `pip install -e ".[dev]"`. So as configured today, CI
always exercises the regex fallback path, never this function — which is
likely why this bug went unnoticed.

Because of that, the new tests call `_format_date_like_original` directly
rather than through `deidentify()`/`shift_dates()`, so they actually exercise
the changed code regardless of whether `python-dateutil` happens to be
installed in the environment running them.

Separately, it might be worth adding `python-dateutil` to the `dev` extra (or
a CI matrix leg) so this path gets real coverage going forward — happy to
open that as its own small PR if useful, didn't want to bundle a new
test-time dependency into this fix.

## Testing
- [x] I have tested this change with different models/inputs

Ran:
- `pytest tests/unit/test_pii.py -q` → 119 passed
- `pytest tests/ -q` (full suite) → 1701 passed, 22 skipped, 0 failed

## Documentation
- [x] I have added docstrings to new functions/classes

## Code Quality
- [ ] For Swift/OpenMedKit changes, I ran `make format-swift` and `make lint-swift`
- [x] I have commented my code, particularly in hard-to-understand areas

## Dependencies

## Checklist
- [x] I have read the contributing guidelines

## Related Issues
No existing issue tracks this; opening the PR directly since it's a small,
self-contained fix. Follow-up: [#604](https://github.com/maziyarpanahi/openmed/pull/604)
fixed an unrelated zero-day-shift bug in the same area.

## Screenshots/Examples
Not applicable (backend logic fix, no UI surface).


maziyarpanahi:
Thanks for the two-digit-year date-format fix. I reviewed the PR and the related date-shifting paths; there is no linked issue, so there were no issue labels to copy.

I added a follow-up commit to cover the default/offline path as well as the dateutil formatting helper:
- widened the no-dateutil `_shift_date_basic()` slash/dash trailing-year patterns to accept 2-4 digit years
- added public `_shift_date()` regressions for slash and dash two-digit-year inputs
- added a forced no-dateutil test so the repo default/dev install path preserves separator and field order too

Validation:
- `.venv/bin/python -m pytest tests/unit/test_pii.py -q` (121 passed)
- `PATH="$PWD/.venv/bin:$PATH" make format-check`
- `PATH="$PWD/.venv/bin:$PATH" make lint`
- GitHub CI passed on the updated PR head after approving the held fork workflow run.

maziyarpanahi (approved):
Approved after final review: CI is green and merge state is clean.

PR #1002, opened 2026-06-28
Build streaming incremental token-classification inference

### LINKED ISSUE #962: Build streaming incremental token-classification inference with bounded tail latency
## Summary
Long clinical notes arrive as growing buffers (dictation, HL7 feeds), but the current pipeline re-runs the whole document on every append, so latency grows quadratically and the UI stalls. A correct streaming path must re-decode only the affected window while preserving span boundaries that may straddle the previously emitted prefix, which is architecturally hard because subword offsets, Viterbi back-pointers, and overlap suppression all assume a fixed input.

## Scope
- [ ] Add a StreamingTokenClassifier that maintains a rolling window with a committed prefix and a recomputation tail sized by max entity span length plus tokenizer context
- [ ] Make the Viterbi decoder incremental: persist back-pointer and score state at the commit boundary and resume decoding from there instead of from token zero
- [ ] Define an emit/retract protocol so spans that extend past the prior commit point are retracted and re-emitted exactly once with stable ids
- [ ] Wire offset reconciliation so byte offsets in retract/emit events map back to absolute document positions across windows
- [ ] Expose an async generator API on the MLX and Torch pipelines and a service endpoint that streams events without buffering PHI to disk
- [ ] Add a deterministic replay harness that feeds documents one character at a time and one chunk at a time and asserts identical final spans

## Acceptance criteria
- [ ] Final emitted spans are byte-identical to a full-document single-shot run across the streaming golden corpus (zero span diff)
- [ ] p99 incremental-append latency is bounded and independent of document length up to 50k tokens (within 1.5x of single-window latency)
- [ ] Entity recall on the streaming eval set is within 0.2 points of the batch baseline with zero new leakage events
- [ ] Each entity id is emitted at most once after the stream completes and retraction events reconcile to the final set
- [ ] No PHI is written to disk or logs during streaming; verified by a redaction-audit test on the event log

## Out of scope
- Generative PII model streaming (token-classification only)
- Cross-document coreference or session-level entity linking
- Network transport tuning beyond the existing service endpoint

## Files
- openmed/core/decoding/viterbi.py
- openmed/core/decoding/spans.py
- openmed/processing/advanced_ner.py
- openmed/mlx/inference.py
- openmed/service/app.py
- openmed/service/runtime.py

---
Task: OM-599  ·  Milestone: v1.8  ·  Priority: P0  ·  Size: XL
Depends on: —  ·  Blocks: —
Roadmap: v1.8 hard initiative
Spec: PLANS/V2/EXECUTION/tasks/OM-599.md

maziyarpanahi:
Closes #962.

Adds bounded incremental token-classification streaming with resumable Viterbi state, stable emit/retract/final span events with absolute byte offsets, backend async generators for MLX and Torch, and an NDJSON PII extraction stream endpoint with PHI-safe audit payloads. Tests cover replay parity, retractions, bounded window latency, audit leakage, OpenAPI/client parity, and the full suite is green.


maziyarpanahi:
Thank you @maziyarpanahi. I reviewed this against #962 and the current `master`, then validated the branch with the relevant local checks and hosted required checks.

Validation:
- GitHub required checks are passing for HEAD `cbfe9f26`.
- Focused local validation was run for the touched surface before this merge pass.

This satisfies the linked issue and is ready to merge.

PR #1181, opened 2026-07-03
fix: correct off-by-one span offsets in multilingual PII regression tests

### LINKED ISSUE #1192: Fix multilingual PII regression fixture span offsets
## Problem

The multilingual PII regression fixtures include stale mock entity offsets. On current master, running:

```bash
.venv/bin/python -m pytest tests/unit/test_pii_multilingual_regression.py -q -W always
```

reproduces span-validation warnings from `tests/unit/test_pii_multilingual_regression.py`. Some Unicode-normalized Hindi and Telugu fixtures can also remap to valid but incorrect substrings, so label-only assertions miss the fixture bug.

## Expected behavior

The mocked entity offsets should match the detection text sent to the analyzer, the extracted entity text should match the expected fixture substring, and the regression file should run without `SpanValidationWarning` noise.

## Resolution

Tracked by PR #1181.

AlexFucuson9:
## Summary

Fixes incorrect mock entity offsets in the multilingual PII regression fixtures and adds exact entity-text assertions for the affected rows.

## Problem

The regression file emitted `SpanValidationWarning` warnings on current `master` because several mocked entity spans were stale. Some Hindi and Telugu fixtures also used original-text coordinates where the mocked model output needs detection-text coordinates, which can remap to valid but incorrect substrings while label-only assertions still pass.

## Fixes

- Corrects the original off-by-one fixture offsets in French, Italian, Spanish, and Dutch cases.
- Computes mock spans from the detection text used by the analyzer, with compatibility for the stale PR branch and current `master`.
- Tightens affected assertions so extracted entity text must match the expected fixture substring.
- Covers Hindi and Telugu normalized-text cases that were previously able to pass with semantically wrong remapped spans.

## Verification

```bash
.venv/bin/python -m pytest tests/unit/test_pii_multilingual_regression.py -q -W error::openmed.core.quality_gates.SpanValidationWarning
uv run ruff check tests/unit/test_pii_multilingual_regression.py
uv run ruff format --check tests/unit/test_pii_multilingual_regression.py
```

Before the fix, current `master` reproduced 7 `SpanValidationWarning` warnings from this file. After the fix, the focused test passes with span warnings promoted to errors.

Fixes #1192

### DIFF
--- a/tests/unit/test_pii_multilingual_regression.py
+++ b/tests/unit/test_pii_multilingual_regression.py
@@ -16,6 +16,11 @@ import pytest
 from openmed.core.pii import extract_pii
 from openmed.processing.outputs import EntityPrediction, PredictionResult
 
+try:
+    from openmed.core.script_detect import normalize_for_pii_detection
+except ImportError:  # pragma: no cover - compatibility with stale PR branches
+    normalize_for_pii_detection = None
+
 # ---------------------------------------------------------------------------
 # Helpers
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


vanthinh6886:
Thanks for the review @maziyarpanahi.

You're right that there's no maintainer-created issue for this. I discovered the 16 `SpanValidationWarning` warnings while running the test suite locally (`pytest tests/test_pii_multilingual_regression.py -v`) and traced them to off-by-one fixture offsets.

To clarify the scope: this PR only touches test fixture data — no production code is changed. The fixes correct mock entity spans in `_ent()` helpers so that `text[start:end]` actually matches the expected entity text. Before: 16 warnings. After: 0 warnings from this file.

If you'd prefer I wait for a maintainer-tracked issue before proceeding, totally understand. Or if you'd like to open one, I can link this PR to it. The fix is narrow (1 file, 12 lines) and ready to merge whenever the scope is confirmed.

PR #1188, opened 2026-07-04
Add redacted-PDF text-layer leakage and visual fidelity verification

### LINKED ISSUE #1101: Add redacted-PDF text-layer leakage and visual fidelity verification
## Summary
Drawing black boxes on a PDF (OM-182 re-render) hides PHI visually but a copy-pasteable text layer underneath frequently survives, which is a classic real-world redaction failure (selectable text leaks the very names that look blacked out). The re-render tooling has no verifier proving the text layer was actually scrubbed and that the visible pixels changed. We need a fidelity check that fails closed when residual text or unchanged pixels remain under a redaction box.

## Scope
- [ ] Create openmed/multimodal/verify_pdf.py that, given an original PDF, its redaction spans/bboxes, and the redacted output, extracts the output's text layer and asserts no redacted span text remains selectable.
- [ ] Render the redacted regions to pixels and verify they visibly changed versus the original (so a box was actually drawn, not just text removed or vice versa).
- [ ] Produce a structured fidelity report: per-region residual-text findings, per-region visual-change check, and an overall pass/fail.
- [ ] Expose verify_redacted_pdf(original, redacted, spans) and a CLI subcommand, returning a non-zero exit / raising on residual leakage.
- [ ] Cover the two canonical failure modes in tests: text-layer leak with visual box present, and visual box missing with text removed.

## Acceptance criteria
- [ ] verify_pdf.py detects residual selectable PHI text under a redaction box on a crafted leaky-redaction fixture and reports failure.
- [ ] It detects an unchanged-pixel region (no visible box drawn) on a crafted fixture and reports failure.
- [ ] A correctly redacted PDF (text scrubbed and pixels changed) passes with an all-clear report.
- [ ] The fidelity report is structured per-region and feeds the audit report.
- [ ] verify_redacted_pdf raises / exits non-zero when any residual leakage is found.
- [ ] .venv/bin/python -m pytest tests/ -q

## Out of scope
- OCR-based residual detection on scanned (image-only) PDFs (covered by image re-OCR in OM-065/OM-075)
- Performing the redaction itself (OM-182)

## Files
- openmed/multimodal/verify_pdf.py
- openmed/cli/verify_pdf.py
- tests/unit/multimodal/test_verify_pdf.py
- tests/unit/multimodal/fixtures/leaky_redaction.pdf
- tests/unit/multimodal/fixtures/clean_redaction.pdf

---
Task: OM-222  ·  Milestone: v1.8  ·  Priority: P1  ·  Size: M
Depends on: OM-060, OM-182  ·  Blocks: —
Roadmap: v1.7 expansion
Spec: PLANS/V2/EXECUTION/tasks/OM-222.md

maziyarpanahi:
Closes #1101.

Drawing a black box over PHI hides it visually, but a copy-pasteable text layer underneath frequently survives — a classic redaction failure where selectable text still leaks the names that look blacked out. This adds a verifier that fails closed on that.

## What
- `openmed.multimodal.verify_redacted_pdf(original, redacted, spans)` checks, per region: (a) no selectable word remains under the redaction box (text-layer leak), and (b) an opaque redaction box actually covers the region (a silently-unapplied redaction shows no visible change). When a raster backend is available it also renders each region to pixels and requires them to differ; otherwise it falls back to vector-content checks.
- A structured, PHI-safe `PdfFidelityReport` (per-region offsets, bounding boxes, and residual-text hashes — never plaintext) suitable for embedding in the audit report, with `raise_for_leakage()` / `strict=True`.
- An `openmed verify-pdf` CLI subcommand that prints the report and exits non-zero on residual leakage.

## Tests
- Deterministic cases (run without the multimodal extra via a faked `pdfplumber`): residual text under a box, an unchanged region with no box, a correctly redacted region, the pixel path via an injected rasterizer, PHI-safe report, char-span resolution, and strict raising.
- An end-to-end test with real `pdfplumber` over synthetic generated PDFs (leaky fails, clean passes, an unchanged copy fails). Full suite green.

Redaction itself is out of scope (that is the re-render task); this only verifies a provided redacted PDF. Scanned image-only PDFs are covered by image re-OCR elsewhere.

### DIFF
--- a/openmed/cli/main.py
+++ b/openmed/cli/main.py
@@ -36,6 +36,7 @@ from ..core.policy import CANONICAL_POLICY_NAMES, canonical_policy_name
 from .active_learning import add_active_learning_command
 from .calibrate import add_calibrate_command
 from .gates import add_gates_command
+from .verify_pdf import add_verify_pdf_command
 
 _ANALYZE_TEXT = None
 _GET_MODEL_MAX_LENGTH = None
@@ -186,6 +187,7 @@ def build_parser() -> argparse.ArgumentParser:
     _add_doctor_command(subparsers)
     add_calibrate_command(subparsers)
     add_gates_command(subparsers)
+    add_verify_pdf_command(subparsers)
     return parser
 
 
--- a/openmed/cli/verify_pdf.py
+++ b/openmed/cli/verify_pdf.py
@@ -1,0 +1,121 @@
+"""CLI subcommand for redacted-PDF text-layer and visual fidelity verification.
+
+Wraps :func:`openmed.multimodal.verify_pdf.verify_redacted_pdf`. Given an
+original PDF, a redacted PDF, and the redaction regions, it prints a PHI-safe
+fidelity report and exits non-zero when any region still leaks selectable text
+or shows no visible redaction box.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+from pathlib import Path
+from typing import Any, Sequence, TextIO
+
+
+def add_verify_pdf_command(subparsers: "argparse._SubParsersAction") -> None:
+    """Register the ``verify-pdf`` subcommand on the top-level parser."""
+    parser = subparsers.add_parser(
+        "verify-pdf",
+        help="Verify a redacted PDF scrubbed its text layer and drew redaction boxes.",
+    )
+    parser.add_argument("original", type=Path, help="Path to the original PDF.")
+    parser.add_argument("redacted", type=Path, help="Path to the redacted PDF.")
+    parser.add_argument(
+        "--spans",
+        type=Path,
+        required=True,
+        metavar="FILE",
+        help=(
+            "JSON file listing the redacted regions: a list of "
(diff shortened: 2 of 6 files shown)


maziyarpanahi:
Hardened after an adversarial self-review. Fixes in this revision:
- **Fail-closed on unreadable input:** a missing/unreadable original PDF no longer silently reports a clean pass — visual verification that hits an I/O error now surfaces it (`FileNotFoundError`) instead of falling back to a pass.
- **Stroke-only rectangles are no longer treated as redaction boxes:** `_is_opaque` now requires an actual fill, so table borders/rules (which pdfplumber still reports a `non_stroking_color` for) can't satisfy the box check.
- `(page, bbox)` tuples are accepted as regions (previously raised `TypeError`).
- The report no longer echoes raw file paths (which can contain identifiers), keeping it PHI-safe.
- `summary()` counts are now disjoint.
- Added regression tests for each of the above plus full CLI coverage (exit codes 0/1/2, `--output`, object-wrapped spans). Suite: 15 verify-pdf tests green; full suite green.

Issue #1364, opened 2026-07-05
OM-800: Per-request resource and timeout budgets with cooperative cancellation


maziyarpanahi:
**Task: OM-800**

## Summary

Add opt-in per-request resource and timeout budgets with cooperative cancellation to the extraction/de-identification path. Long or pathological inputs, or a stalled pipeline stage, can currently run unbounded. Callers need a way to bound the wall-clock time and input size of a single de-identification/extraction request and receive a clean, typed error when a budget is exceeded — without hard thread kills or leaving partial state behind.

## Motivation

- On-device and local-first deployments must stay responsive; a single oversized note should not hang a request indefinitely.
- Batch jobs need a defensive per-item bound so one bad row cannot stall the whole run.
- Cancellation must be cooperative (checked at safe checkpoints between pipeline stages / batch items) so the library never corrupts partial output or raw-PHI state.

## Scope

- A per-request budget object carrying an optional wall-clock (max-time) budget and an optional max-input-length (character) budget.
- A typed error raised when a budget is exceeded, distinct from arbitrary runtime errors.
- Deadline checks inserted at safe checkpoints:
  - between pipeline stages in `Pipeline.run` (`openmed/core/pipeline.py`),
  - between batch items / the batch extraction loop in `openmed/core/pii.py` (`_extract_pii_batch`, `_deidentify_batch`),
  - between chunks/items in the `BatchProcessor` loop (`openmed/processing/batch.py`).
- The input-length guard is enforced up front (before model inference) against the max-input-length budget.
- Budgets are threaded through `extract_pii` and `deidentify` as an optional parameter.

## Acceptance Criteria

- [ ] A `RequestBudget` (or equivalent) type carries `max_wall_time` (seconds) and `max_input_chars`, both optional; `None` means unlimited.
- [ ] A typed `BudgetExceededError` (subclass of a suitable stdlib base) is raised when either budget is exceeded. The error message and any attributes contain NO raw PHI / input text — only counts, limits, and the stage/checkpoint name.
- [ ] Cancellation is cooperative: the deadline is checked at safe checkpoints between pipeline stages and between batch items — never via thread kill or signal-based interruption.
- [ ] `extract_pii(...)` and `deidentify(...)` accept an optional `budget=` argument; when omitted or `None`, behavior is byte-for-byte identical to today (fully backward compatible).
- [ ] The max-input-length guard rejects over-length input before running inference.
[...]

PR #1373, opened 2026-07-05
OM-800: Per-request resource and timeout budgets with cooperative cancellation

maziyarpanahi:
## Status

Closed unmerged during the post-v2 reconciliation. This is an archival implementation associated with issue #1364. The head branch was later recreated, so this pull request cannot be reopened; current work continues from the latest master on the same named branch.


maziyarpanahi:
Closing this stale implementation during the post-v2.0 reconciliation. It is roughly 1,000 commits behind current master and no longer merges cleanly. The associated issue #1364 remains open; this does not mark the work complete. Any v2.1 implementation should restart from current master after scope and dependencies are reconfirmed.

PR #1390, opened 2026-07-05
fix: replace bare asserts with explicit exceptions (P0 — python -O survival)

AlexFucuson9:
## Summary

Replace bare `assert` statements with explicit exceptions that survive `python -O` (optimized mode).

## Problem

Bare `assert` statements are **silently stripped** when Python runs with `-O` flag. Critical validation checks in Pydantic validators and auth code would disappear, allowing invalid data through.

## Changes

| File | Line | Fix |
|------|------|-----|
| `openmed/service/schemas.py` | 374 | `assert normalized is not None` -> `if ... raise ValueError` |
| `openmed/service/schemas.py` | 741 | Same pattern (v2 validator) |
| `openmed/service/auth.py` | 747 | `assert exp is not None` -> `if ... raise ValueError` |

## Impact

- **P0**: All 3 asserts validate critical healthcare data (confidence thresholds, JWT tokens)
- Under `python -O`, invalid confidence values and malformed tokens would silently pass validation
- Explicit `ValueError`/`AssertionError` replacements always execute regardless of optimization level

## Testing

All files pass `python3 -m py_compile` verification.

### DIFF
--- a/openmed/service/auth.py
+++ b/openmed/service/auth.py
@@ -744,7 +744,8 @@ def _validate_jwt_claims(
     now: float,
 ) -> None:
     exp = _numeric_claim(payload, "exp", required=True)
-    assert exp is not None
+    if exp is None:
+            raise ValueError("Token expiration (exp) claim must be present")
     if now > exp + config.jwt_leeway_seconds:
         raise invalid_credentials()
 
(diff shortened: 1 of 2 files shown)


AlexFucuson9:
This PR is superseded by #1404 which has been merged. Closing.

PR #1391, opened 2026-07-05
fix: add explicit UTF-8 encoding to Path.read_text() in security scripts (P1)

AlexFucuson9:
## Summary

Add explicit `encoding="utf-8"` to `Path.read_text()` calls in security scripts.

## Problem

`Path.read_text()` without encoding defaults to system locale (cp1252 on Windows), causing silent data corruption when reading UTF-8 content like JSON and TOML config files.

## Changes

`scripts/security/pip_audit_gate.py`:
- `path.read_text()` -> `path.read_text(encoding="utf-8")` (2 instances)
- Both read JSON/TOML files which are UTF-8 by spec (RFC 8259 / TOML spec)

## Testing

File passes `python3 -m py_compile`.

Fixes #1409

PR #1396, opened 2026-07-06
fix: add explicit UTF-8 encoding to write_text in vulnerability_scan_gate (P1)

### LINKED ISSUE #1411: bug: Path.write_text() without explicit encoding in vulnerability_scan_gate corrupts JSON on Windows
## Description

`scripts/security/vulnerability_scan_gate.py` calls `Path.write_text()` without specifying `encoding="utf-8"` in `write_summary()`:

```python
path.write_text(json.dumps(payload, indent=2, sort_keys=True) + "\n")
```

On Windows with non-UTF-8 default locale, this writes the JSON artifact using the platform encoding (e.g. `cp1252`). If `GateResult` contains vulnerability descriptions or package names with non-ASCII characters, the output will be silently corrupted or raise `UnicodeEncodeError`.

This is a CI gate script -- corrupted JSON artifacts can cause downstream tooling to misparse results or fail silently.

## Expected Behavior

All `Path.write_text()` calls that produce JSON artifacts should explicitly pass `encoding="utf-8"` for cross-platform correctness and interoperability.

## Environment

- All platforms, especially Windows with non-UTF-8 locale
- Python 3.11+

## Proposed Fix

Add `encoding="utf-8"` to the `write_text()` call in `write_summary()`. PR #1396 implements this fix.

AlexFucuson9:
## Problem

`Path.write_text()` without explicit encoding defaults to the system locale (`cp1252` on Windows), causing silent data corruption when writing UTF-8 JSON content.

PR #1391 fixed the `read_text()` calls in `pip_audit_gate.py`. This PR fixes the remaining `write_text()` call in `vulnerability_scan_gate.py`.

## Change

`scripts/security/vulnerability_scan_gate.py`:
- `path.write_text(json.dumps(...) + "\n")` → `path.write_text(json.dumps(...) + "\n", encoding="utf-8")`

The payload is JSON (RFC 8259 = UTF-8), so the encoding must be explicit.

## Testing

File passes `python3 -m py_compile`.

Fixes #1411

### DIFF
--- a/scripts/security/vulnerability_scan_gate.py
+++ b/scripts/security/vulnerability_scan_gate.py
@@ -287,7 +287,7 @@ def write_summary(path: Path, result: GateResult) -> None:
     """Write the gate result as a JSON artifact."""
     path.parent.mkdir(parents=True, exist_ok=True)
     payload = asdict(result)
-    path.write_text(json.dumps(payload, indent=2, sort_keys=True) + "\n")
+    path.write_text(json.dumps(payload, indent=2, sort_keys=True) + "\n", encoding="utf-8")
 
 
 def print_blocking_findings(findings: list[Finding]) -> None:


AlexFucuson9:
Created issue #1411 as the maintainer-backed issue for this fix. The `write_text()` in `vulnerability_scan_gate.py` indeed lacks explicit encoding, which corrupts JSON artifacts on Windows.

Fixes #1411

PR #1404, opened 2026-07-07
fix: replace bare asserts with explicit exceptions in service layer

### LINKED ISSUE #1407: bug: assert statements in service layer bypassed under python -O, breaking JWT validation
## Description

Production code in `openmed/service/auth.py`, `openmed/service/schemas.py`, and `openmed/mcp/server.py` uses `assert` statements for invariant guards. Under `python -O` (optimize mode), all `assert` statements are stripped entirely, causing:

1. **Security-critical**: `_validate_jwt_claims` uses `assert exp is not None` before checking token expiry. Under `-O`, this guard vanishes — the expiry comparison receives `None` and either crashes with `TypeError` or silently bypasses the check.

2. **Runtime crash**: `openmed/service/schemas.py` uses `assert normalized is not None` in two Pydantic validators for `confidence_threshold` — these become no-ops under `-O`.

3. **Unguarded env parsing**: `openmed/mcp/server.py` uses bare `int(os.getenv(...))` which crashes with `ValueError` on malformed env vars (e.g. `OPENMED_MCP_PORT=abc`).

## Steps to Reproduce

```bash
python -O -c "
from openmed.service.auth import _validate_jwt_claims
# JWT with missing exp claim will bypass expiry check
"
```

## Expected Behavior

Invariant checks should survive `python -O` and use explicit `if ... raise` patterns.

## Environment

- Python 3.11+
- All platforms

## Proposed Fix

Replace bare `assert` with explicit `if condition: raise ...` and add `_safe_int_env()` helper for env var parsing.

---

*Sourced from code review — PR #1404 implements the fix.*

AlexFucuson9:
## Description

Replace `assert` statements in production code with explicit `if ... raise` checks so invariant guards survive `python -O` (which strips all `assert` statements).

**Fixes #1407**

## Change Type

- [x] Bug fix (non-breaking change that fixes an issue)

### `openmed/service/auth.py` — JWT expiry guard (security-critical)

`_validate_jwt_claims` used `assert exp is not None` to guard the JWT `exp` claim before checking token expiry. Under `python -O`, this assert is stripped entirely — meaning the `now > exp + leeway` comparison would receive `None` and crash with `TypeError`, or worse, the expiry check could be bypassed in edge cases.

**Fix:** `if exp is None: raise invalid_credentials()` — explicit, `-O`-safe, and semantically identical.

### `openmed/service/schemas.py` — Pydantic validators (2 instances)

Two `@field_validator` / `@validator` methods for `confidence_threshold` used `assert normalized is not None` after calling `_normalize_confidence_threshold()`. While the current code path makes `None` unlikely (input is typed `float`, not `Optional`), the assert serves only as a type-narrowing hint for mypy/pyright and vanishes under `-O`.

**Fix:** `if normalized is None: raise ValueError(...)` — explicit, `-O`-safe.

### `openmed/mcp/server.py` — unguarded `int(os.getenv())` (crash on malformed input)

Two call sites used `int(os.getenv("OPENMED_MCP_PORT", "8081"))` which crashes with `ValueError` if the env var contains non-numeric text (e.g. `OPENMED_MCP_PORT=abc`).

**Fix:** Add `_safe_int_env(name, default)` helper that catches `ValueError`/`TypeError` and falls back to the default.

## Tests Run

```
tests/unit/service/ — 191 passed (excluding pre-existing gRPC stubs failure)
tests/unit/mcp/ — included above
Full suite — 2574 passed, 45 skipped
```

## Linked Issue

Fixes #1407

### DIFF
--- a/openmed/mcp/server.py
+++ b/openmed/mcp/server.py
@@ -27,6 +27,18 @@ from openmed.utils.validation import validate_model_name
 
 RuntimeProvider = Callable[[], ServiceRuntime]
 
+
+def _safe_int_env(name: str, default: int) -> int:
+    """Read an environment variable as an int, falling back to *default* on error."""
+    raw = os.getenv(name)
+    if raw is None:
+        return default
+    try:
+        return int(raw)
+    except (TypeError, ValueError):
+        return default
+
+
 MCP_INSTRUCTIONS = (
     "OpenMed exposes local clinical NLP, PII extraction, and de-identification "
     "tools. Use synthetic examples for tests and docs. Only send real PHI to "
@@ -532,7 +544,7 @@ def create_mcp_server(
         instructions=MCP_INSTRUCTIONS,
         website_url="https://openmed.life/docs/",
         host=host or os.getenv("OPENMED_MCP_HOST", "127.0.0.1"),
-        port=port or int(os.getenv("OPENMED_MCP_PORT", "8081")),
+        port=port or _safe_int_env("OPENMED_MCP_PORT", 8081),
         streamable_http_path=streamable_http_path,
         stateless_http=True,
         json_response=True,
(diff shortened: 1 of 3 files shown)


AlexFucuson9:
Hi @maziyarpanahi — the associated issue is now filed:

**Fixes #1407** — describes the three concrete problems (JWT validation bypass under `python -O`, stripped asserts in Pydantic validators, and crash on malformed env vars) with reproduction steps.

I've updated the PR description accordingly. Happy to adjust scope or add acceptance criteria if needed.

PR #1547, opened 2026-07-11
feat: add full-width/half-width normalization with offset preservation

### LINKED ISSUE #1468: Implement full-width/half-width (全角/半角) normalization with NFKC-aware offset preservation for Chinese text
## Summary
Chinese documents freely mix full-width Latin letters, digits, and punctuation (ＡＢＣ１２３，：) with half-width ASCII, so a phone number or ID can appear as full-width digits that regex PHI patterns and tokenizers never match. The hard part is that blunt NFKC normalization can change string length and merge characters, silently shifting every downstream PHI offset, so OpenMed needs width normalization that records exactly which code points changed and stays reversible for redaction.

## Scope
- [ ] Add a normalize_width function in openmed/processing/zh_normalize.py that maps full-width ASCII variants (U+FF01-FF5E) to half-width and the ideographic space U+3000 to a configurable target, while preserving genuinely full-width CJK content untouched.
- [ ] Return an offset alignment alongside the normalized text so that since width normalization is almost always 1:1 in code points, offsets map straight through, with explicit handling and tests for the rare NFKC many-to-one cases.
- [ ] Add a policy switch to choose the recommended CJK convention (Latin/symbols/digits to half-width, everything else to full-width) versus strict NFKC, surfaced via openmed/core/config.py.
- [ ] Wire width normalization as an early pre-pass so existing Latin-oriented PHI regexes in openmed/core/pii_i18n.py LANGUAGE_PII_PATTERNS match full-width phone/ID inputs after normalization, with spans mapped back to original code points.
- [ ] Extend openmed/core/decoding/spans.py whitespace trimming to recognize U+3000 so CJK spans are trimmed correctly.
- [ ] Add tests with synthetic full-width clinical identifiers verifying both detection-after-normalization and exact original-span recovery.

## Acceptance criteria
- [ ] A synthetic full-width phone number （１３８１２３４５６７８） is matched by the existing half-width-oriented PHI pattern after normalization, and its redaction span maps back to the exact original full-width code points.
- [ ] Offset-correctness gate: for 500 synthetic strings, every normalized offset round-trips to the correct original code-point offset, including inputs containing U+3000.
- [ ] Genuinely full-width CJK characters (Han, full-width quotes) are left unchanged by normalize_width.
- [ ] The CJK-convention policy converts Latin/digits/symbols to half-width while keeping Han full-width, matching documented expected output on a probe set.
- [ ] trim_span_whitespace treats U+3000 as trimmable whitespace and does not strip interior Han characters.

## Out of scope
- Simplified/Traditional conversion (separate task).
- Pinyin or numeral conversion.

## Files
- openmed/processing/zh_normalize.py
- openmed/core/pii_i18n.py
- openmed/core/decoding/spans.py
- openmed/core/config.py

---
Task: OM-642  ·  Milestone: v1.9  ·  Priority: P1  ·  Size: M
Depends on: —  ·  Blocks: —
Roadmap: Chinese & Indian languages initiative
Spec: PLANS/V2/EXECUTION/tasks/OM-642.md

pardeep-singh:
Implements #1468 (OM-642). Chinese documents mix full-width Latin letters, digits, and punctuation (`ＡＢＣ１２３，：`) with half-width ASCII, so a phone number or ID in full-width digits never matches the half-width-oriented PHI regexes. Blunt NFKC can change string length and silently shift every downstream PHI offset. This adds width normalization that records exactly which code points changed and maps detected spans back to the original.

## Changes

- **`openmed/processing/zh_normalize.py`** (new)
  - `normalize_width(text, *, convention="cjk", space_target=" ")` -> `WidthNormalization` with a per-character offset alignment (`char_origins`) and `to_original_span(start, end)`. Full-width ASCII (U+FF01-FF5E) -> half-width; U+3000 -> configurable target; genuine full-width CJK (Han, quotes) left untouched. The rare NFKC many-to-one expansions (`㎏` -> `kg`) are handled by mapping every normalized char back to its source code point.
  - `detect_width_normalized(text, matcher, ...)` -- the early pre-pass: runs a caller-supplied matcher on the normalized text and remaps every span back to original offsets, so existing PHI patterns match full-width inputs.
- **`openmed/core/config.py`** - `cjk_width_convention` policy switch (`cjk` recommended convention vs strict `nfkc`), validated in `__post_init__`.

## Acceptance criteria

- [x] A full-width identifier is matched by the existing half-width-oriented PHI engine after normalization, and its span maps back to the exact original full-width code points (demonstrated end-to-end with `find_semantic_units` on a full-width date, and offset mapback on a full-width phone).
- [x] Offset-correctness gate: for 500 synthetic strings (including U+3000), every normalized offset round-trips to the correct original code-point offset.
- [x] Genuine full-width CJK characters are left unchanged.
- [x] The CJK-convention policy converts Latin/digits/symbols to half-width while keeping Han full-width.
- [x] `trim_span_whitespace` treats U+3000 as trimmable and does not strip interior Han.

## Scope notes (files vs the issue list)

- **`openmed/core/decoding/spans.py` -- no change needed.** `trim_span_whitespace` already uses `str.isspace()`, and `"　".isspace()` is `True`, so the ideographic space is already trimmed correctly. This is verified and locked with tests (trims U+3000, keeps interior Han) rather than duplicating the logic.
- **`openmed/core/pii_i18n.py` -- not modified.** The width pre-pass is provided as a composable helper in the processing layer (`detect_width_normalized`) that consumes the existing detection engine (`find_semantic_units`) and `get_patterns_for_language`, rather than modifying core detection. This avoids an import cycle (`core.config`/`core.pii_i18n` -> `processing`) and keeps the change minimal. Happy to inline it into `pii_i18n.py` if you would prefer the wiring to live there.

## Testing

- New `tests/unit/processing/test_zh_normalize.py` (13 tests): width mapping, ideographic-space handling, offset round-trip gate (500 strings), NFKC many-to-one, pre-pass detection via the existing engine, U+3000 trimming, and the config switch.
- Full offline unit suite: `4004 passed, 58 skipped`. `ruff check` / `format` clean (0.15.20). No new dependencies (stdlib `unicodedata`).

## Out of scope (per the issue)

- Simplified/Traditional conversion.
- Pinyin or numeral conversion.

Closes #1468


pardeep-singh:
Green and mergeable against current master, and independent of my other open PRs. This adds full-width/half-width normalization with an exact offset alignment (500-string round-trip gate), stdlib-only. Two Files-list items needed no change and are explained in the description (`decoding/spans.py` already trims U+3000 via `str.isspace()`; the width pre-pass composes the existing `find_semantic_units` engine to avoid a `core -> processing` import cycle). Whenever you have review bandwidth, a review would be appreciated.

maziyarpanahi:
Thank you @pardeep-singh — the offset-preserving normalization and synthetic round-trip coverage were a strong base.

I updated the branch to current `master` and added maintainer follow-up commit `a6663eff` to connect the helper to the production PII detection path, compose it with the existing Unicode defense, honor the configured CJK/NFKC policy, preserve original spans through semantic matching, complete config serialization, and document the setting.

Validation:
- `4,678 passed, 52 skipped`
- `make lint`
- `make format-check`
- `make type-check`
- `make docs-build` (strict)
- all 18 hosted checks passed, including amd64/arm64 container smoke builds

This now satisfies #1468 and is ready to merge.

PR #1614, opened 2026-07-16
fix(clinical): scope experiencer cues across contrastive clauses (#277)

### LINKED ISSUE #277: Add the experiencer axis on top of the ConText engine
## Summary
OM-041 must determine the experiencer of a clinical span (patient vs other person) so family-history conditions are NOT recorded as the patient's ('mother had breast cancer' must be excluded from the patient Condition list). The epic bundles experiencer with three other axes. This is the narrow, grabbable experiencer decision layer on the shared ConText engine, and it directly protects record correctness and re-identification posture (family mentions are quasi-identifiers).

## Scope
- [ ] Add resolve_experiencer(span, modifier_hits) -> 'patient'|'other' to clinical/context.py.
- [ ] Cover family/other-person cues ('mother', 'father', 'sister', 'family history of', 'FHx', 'her husband', 'the donor') from the ConText experiencer lexicon, with scope so only concepts inside the family clause are marked 'other'.
- [ ] Expose experiencer per span; document that experiencer='other' excludes the span from the patient record (consumed by grounding OM-040).
- [ ] Tests: 'mother had breast cancer' -> other; 'patient has breast cancer' -> patient; a sentence with both a family clause and a patient clause assigns each concept the correct experiencer.

## Acceptance criteria
- [ ] resolve_experiencer marks concepts in family/other-person scope as 'other' and patient concepts as 'patient', including mixed sentences (test).
- [ ] The experiencer result is exposed per span and documented as excluding 'other' spans from the patient record.

## Out of scope
- Negation/uncertainty/temporality axes (sibling tasks).
- Family-history record routing into FamilyMemberHistory resources (grounding/FHIR epic).
- Learned experiencer heads.

## Files
- openmed/clinical/context.py
- tests/unit/clinical/test_experiencer.py

---
Task: OM-112  ·  Milestone: Backlog  ·  Priority: P1  ·  Size: S
Depends on: —  ·  Blocks: —
Roadmap: section 5.2 (lines 616-628); decomposes OM-041
Spec: PLANS/V2/EXECUTION/tasks/OM-112.md

PouyanJay:
## Summary

Addresses #277 (OM-112): the experiencer axis must give each concept in a mixed sentence the correct subject, so a family-history finding is not recorded as the patient's.

## Important: the feature already exists — this fixes the real gap

The issue asks to *add* `resolve_experiencer` to `openmed/clinical/context.py`. During Phase-1 analysis I found that **`resolve_experiencer` already ships** — in `openmed/clinical/experiencer.py` (the v1.9 cue-based experiencer-refinement layer, with `test_experiencer_refine.py` and `docs/clinical/experiencer-refinement.md`). It already returns `patient`/`family`/`other`, covers the issue's cues (mother, father, sister, family history, FHx, husband, donor), and is clause-scoped.

Running the issue's acceptance examples against the shipped code showed it passes every case **except the headline mixed-sentence one**:

| Issue example | Before | After |
| --- | --- | --- |
| `mother had breast cancer` | family | family |
| `patient has breast cancer` | patient | patient |
| `the donor was CMV-positive` | other | other |
| **`Mother had breast cancer, but the patient has hypertension`** → hypertension | **family (wrong)** | **patient** |

The cause: the module bounded a subject cue only at sentence punctuation (`[.!?;]`), so a family cue leaked across a contrastive conjunction into the following patient clause. The issue explicitly requires the mixed family-clause/patient-clause case to work, and points at the ConText engine — whose `scope_terminator_re` already splits on contrastive conjunctions.

## The fix

Add the ConText engine's **contrastive** scope terminators (`but`, `however`, `whereas`) as clause boundaries in `experiencer.py`, so a subject cue cannot reach across them. The coordinating terminators `and`/`or` are **deliberately excluded**: they conjoin findings under one subject ("mother had X **and** Y"), where the family experiencer must still reach the later finding — including `and` would regress that case.

## Deviation from the issue's file list

- The issue lists `context.py` + a new `test_experiencer.py`. Because the shipped `resolve_experiencer` lives in `experiencer.py`, the fix and its tests belong there (`experiencer.py` + the existing `test_experiencer_refine.py`) rather than creating a second, divergent `resolve_experiencer` in `context.py`. This reuses the existing feature instead of duplicating it.

## Tests

New tests cover: the contrastive `but` split (both clause orders), the `however` variant, and an `and`/`or` non-split regression guard so conjoined findings stay with the family.

## Verification

- `.venv/bin/python -m pytest tests/` (the issue's acceptance command): full suite green — 4863 passed, 43 skipped.
- `tests/unit/clinical/`: 607 passed. `pre-commit` clean on the changed files.

Out of scope, per the issue: the negation/uncertainty/temporality axes, FHIR FamilyMemberHistory routing, and learned experiencer heads.

Closes #277

### DIFF
--- a/docs/clinical/experiencer-refinement.md
+++ b/docs/clinical/experiencer-refinement.md
@@ -12,7 +12,7 @@ local subject cues, distinguishing three subjects:
 | Value | Subject | Example cues |
 |---|---|---|
 | `patient` | the patient (default) | none cued |
-| `family` | a relative of the patient | mother, father, sibling, maternal, family history |
+| `family` | a relative of the patient | mother, father, sibling, maternal, family history, FHx |
 | `other` | a non-patient, non-relative subject | donor, roommate, partner, coworker |
 
 This axis is a hard safety boundary for coreference-style aggregation: a
@@ -70,9 +70,12 @@ result = resolve_experiencer(text, span)
 ## Resolution order
 
 1. **Cue** -- the subject cue nearest the span, within the span's clause, wins.
-   Resolution is scoped to the clause (bounded by `.`, `!`, `?`, `;`), so a
-   subject named in a previous sentence does not leak across the boundary. On a
-   tie the more specific `other` subject wins over `family`.
+   Resolution is scoped by sentence punctuation (`.`, `!`, `?`, `;`) and the
+   contrastive subject-switch markers `but`, `however`, and `whereas`, so a
+   subject named in a previous clause does not leak across the boundary.
+   Coordinating `and` and `or` deliberately remain inside the same scope because
+   they commonly join findings about one subject. On a tie the more specific
+   `other` subject wins over `family`.
 2. **Section prior** -- when no cue is found and a `section_experiencer` is
    supplied (for example `family` under a *Family History* heading), it is used.
 3. **Default** -- otherwise the span is attributed to the `patient`.
--- a/openmed/clinical/experiencer.py
+++ b/openmed/clinical/experiencer.py
@@ -58,6 +58,7 @@ EXPERIENCER_REFINEMENT_ADVISORY = (
 # Relatives whose mention attributes a nearby finding to the family.
 _FAMILY_CUES = (
     "family history",
+    "fhx",
     "familial",
     "maternal",
     "paternal",
--- a/openmed/clinical/experiencer.py
+++ b/openmed/clinical/experiencer.py
@@ -108,6 +108,13 @@ _OTHER_CUES = (
 # The coordinating terminators "and"/"or" are deliberately excluded because they
 # usually conjoin findings under the same subject ("mother had X and Y"), where
 # the family experiencer must still reach the later finding.
+#
+# Only the unambiguous contrastive conjunctions are handled here. Broader
+# subject-switching markers are intentionally left out for now: "although",
+# "though", and "yet" are lower-value, and "while" is ambiguous (temporal
+# "while on aspirin" vs contrastive "while the mother ..."), so splitting on it
+# risks regressing temporal clauses. Widening this set is a deliberate
+# follow-up, not an oversight.
 _CLAUSE_BOUNDARY_RE = re.compile(
     r"[.!?;]|(?<!\w)(?:but|however|whereas)(?!\w)",
     re.IGNORECASE,
--- a/openmed/clinical/experiencer.py
+++ b/openmed/clinical/experiencer.py
@@ -102,8 +102,16 @@ _OTHER_CUES = (
     "contact",
 )
 
-# Clause boundaries that a cue may not reach across.
-_CLAUSE_BOUNDARY_RE = re.compile(r"[.!?;]")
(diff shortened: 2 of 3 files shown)


PouyanJay:
@maziyarpanahi there is a something I did here that I would like to have your input on:

Output vocabulary differs from the issue's literal wording. The issue says the function returns 'patient'|'other', and "mother had breast cancer" should be 'other'. The shipped code (and my fix) returns 'patient'|'family'|'other' so that case returns family, not the literal string 'other'. I'm satisfying the issue's intent (non-patient → excluded from the patient record) using the maintainer's existing richer vocabulary, where family is a subtype of "other/non-patient." That's the right call for consistency with shipped code, but it is a deviation from the issue's literal output values and I flagged it in the PR.

PouyanJay:
@maziyarpanahi I also believe a new issue needs to be created to address the following. I did not implement it here not to creep the scope of this issue:

One judgment call I'd value your input on: for the experiencer clause boundary I used only the unambiguous contrastive conjunctions (but, however, whereas) and deliberately left out the coordinating and/or (they conjoin same-subject findings like "mother had X and Y", which must stay family) and the ambiguous while (temporal "while on aspirin" vs contrastive "while the mother…"), since a bare while split would regress temporal clauses.

That's enough to satisfy #277's mixed-clause case, but it's intentionally conservative, markers like although/though/yet aren't handled yet. Do you have a preference on how wide this set should go, or whether an experiencer-specific contrastive lexicon should be shared with the ConText scope_terminators? Happy to broaden it here if you'd like, or to open a separate follow-up issue for it — whichever you prefer. I've documented the current boundary and rationale at _CLAUSE_BOUNDARY_RE so the limit reads as deliberate.

maziyarpanahi:
Both choices are right for this PR.

Please keep the shipped three-value vocabulary: `family` should remain distinct from `other`, with both treated as non-patient by downstream record construction. Collapsing `family` into `other` here would regress the existing public behavior just to match the older issue wording.

Keep the clause-boundary change conservative as well. `but`, `however`, and `whereas` address the acceptance case without breaking same-subject coordination; do not add `and`, `or`, or bare `while` in this PR. `although`/`though`/`yet` can be handled in a separate follow-up with dedicated mixed-clause and temporal regression cases. I would also avoid sharing the generic ConText terminator set in this PR because experiencer scoping has different same-subject semantics.

So no further scope expansion is needed for #277.

PouyanJay:
OK thanks .. this is ready to merge

maziyarpanahi:
Thanks @PouyanJay — I reviewed this against #277 and the conservative clause-boundary choice is the right one. I added one maintainer follow-up commit (`00eb7e0a`) to cover the issue’s explicit `FHx` cue and align the experiencer docs with the implemented punctuation/contrastive-boundary behavior. The focused tests (19), full clinical suite (608 passed, 2 skipped), Ruff, strict docs build, hosted matrix, and fresh current-master merge simulation all pass. I’m merging with your commits preserved.

PR #1644, opened 2026-07-17
Add an eval-result provenance and reproducibility-hash ledger

### LINKED ISSUE #886: Add an eval-result provenance and reproducibility-hash ledger
## Summary
Benchmark results are archived but there is no append-only ledger binding each result to its inputs (model digest, suite version, code hash, seed, config) so a published number can be independently reproduced and audited. Open-benchmark trust requires verifiable provenance.

## Scope
- [ ] Define a ledger record capturing model digest, suite id+version, code hash, seed, env fingerprint, and result hash
- [ ] Append records to a tamper-evident, hash-chained ledger file on each eval run
- [ ] Add a verify command that re-checks the chain and recomputes result hashes

## Acceptance criteria
- [ ] Appending records produces a valid hash chain and tampering with any record fails verification
- [ ] Re-running an identical eval yields a matching result hash, confirming reproducibility
- [ ] Ledger stores only digests and metadata, never raw PHI

## Out of scope
- Cryptographic signing (covered by the existing signed audit report)
- Remote ledger storage

## Files
- openmed/eval/provenance_ledger.py
- tests/unit/eval/test_provenance_ledger.py

---
Task: OM-529  ·  Milestone: v1.9  ·  Priority: P1  ·  Size: M
Depends on: OM-023  ·  Blocks: —
Roadmap: v1.8/v1.9 expansion (post-1.7)
Spec: PLANS/V2/EXECUTION/tasks/OM-529.md

maziyarpanahi:
# Pull Request

## Description

Adds a PHI-safe evaluation provenance ledger that binds aggregate results to model, suite, code, seed, configuration, and environment fingerprints. Records are hash-linked and carry independently recomputable result and reproducibility hashes. Verification checks internal chain consistency and can require an externally trusted head hash to detect a complete chain rewrite.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition/improvement

## Changes Made

- Define a strict, versioned ledger schema with model/config/environment digests, suite identity, code hash, seed, aggregate result metadata, and chain anchors.
- Append each completed evaluation as an atomically persisted hash-linked record, deriving the input reproducibility hash through the existing release hash API.
- Restrict persisted result metadata to schema-controlled keys and numeric, boolean, null, nested aggregate, or content-digest values.
- Add `verify_ledger(..., expected_head_hash=...)` and `verify <ledger> --expected-head sha256:...` so callers can verify against a separately trusted anchor.
- Cover valid chains, partial mutation, deletion, reordering, coherent full-chain rewrites, head mismatch, fail-closed appends, identical reruns, changed inputs, strict PHI-safe metadata, and command success/failure.

## Testing

- [x] I have added tests that prove my feature works
- [x] I have tested this change with different inputs

Validation performed on the rebased head:

- `.venv/bin/python -m pytest tests/unit/eval/ -q` — 685 passed, 5 skipped
- `.venv/bin/python -m pytest tests/ -q` — 6,195 passed, 48 skipped
- `make format`
- `make lint`
- `make format-check`
- `make type-check`
- `pre-commit run --files openmed/eval/provenance_ledger.py tests/unit/eval/test_provenance_ledger.py`
- `make docs-build`
- `python3 -m build`

## Documentation

- [x] I have added docstrings to new functions/classes

No user guide or changelog update is needed for this scoped module; the verification command includes built-in help.

## Code Quality

- [ ] For Swift/OpenMedKit changes, I ran `make format-swift` and `make lint-swift`
- [x] I have commented my code, particularly in hard-to-understand areas

## Dependencies

- [ ] OR I have added new dependencies and they are justified because: ____

## Checklist

- [x] I have read the contributing guidelines
- [x] I have organized my commits appropriately

## Related Issues

Closes #886

Depends on OM-023, completed by #338.

## Screenshots/Examples

Not applicable; this change adds a library API and command-line verifier with no visual interface.

### DIFF
--- a/openmed/eval/provenance_ledger.py
+++ b/openmed/eval/provenance_ledger.py
@@ -1,0 +1,741 @@
+"""Tamper-evident provenance ledger for evaluation results.
+
+The ledger binds aggregate, PHI-safe evaluation results to the exact inputs
+that produced them.  Each record commits to the preceding record, allowing the
+verifier to detect mutation, insertion, deletion, and reordering.  Result
+metadata is intentionally restricted to numbers, booleans, nulls, and content
+digests so raw clinical text cannot be persisted accidentally.
+"""
+
+from __future__ import annotations
+
+import argparse
+import hashlib
+import hmac
+import json
+import math
+import os
+import re
+import sys
+import tempfile
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+from openmed.core.repro_hash import compute_reproducibility_hash
+
+LEDGER_SCHEMA_VERSION = "openmed.eval_provenance_ledger.v1"
+
+_DIGEST_RE = re.compile(r"^sha256:[0-9a-f]{64}$")
+_IDENTIFIER_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._:/+@-]{0,255}$")
+_METADATA_KEY_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9_.-]{0,127}$")
+_SENSITIVE_KEY_TOKENS = frozenset(
+    {
+        "address",
+        "content",
+        "document",
+        "email",
+        "identifier",
+        "mrn",
+        "name",
+        "note",
+        "patient",
+        "phone",
+        "prompt",
+        "raw",
+        "response",
+        "ssn",
+        "text",
+    }
+)
+_AGGREGATE_SUFFIXES = frozenset(
+    {"bytes", "count", "digest", "hash", "length", "ms", "rate", "ratio", "score"}
+)
+
+
+def _canonical_json(value: Any) -> str:
+    return json.dumps(
+        value,
+        ensure_ascii=False,
+        allow_nan=False,
+        separators=(",", ":"),
+        sort_keys=True,
+    )
+
+
+def _hash_json(value: Any) -> str:
+    encoded = _canonical_json(value).encode("utf-8")
+    return f"sha256:{hashlib.sha256(encoded).hexdigest()}"
+
+
+GENESIS_HASH = _hash_json({"schema_version": LEDGER_SCHEMA_VERSION, "sequence": -1})
+
+
+class ProvenanceLedgerError(ValueError):
+    """Base error for malformed or unverifiable provenance ledgers."""
+
+
+class UnsafeResultMetadataError(ProvenanceLedgerError):
+    """Raised when result metadata could persist raw or identifying data."""
+
+
+class LedgerVerificationError(ProvenanceLedgerError):
+    """Raised when an evaluation provenance ledger fails verification."""
+
+
+@dataclass(frozen=True)
+class EvalProvenanceRecord:
+    """One PHI-safe evaluation result and its reproducibility provenance."""
+
+    sequence: int
+    model_digest: str
+    suite_id: str
+    suite_version: str
+    code_hash: str
+    seed: int
+    config_digest: str
+    env_fingerprint: str
+    result_metadata: Mapping[str, Any]
+    result_hash: str
+    reproducibility_hash: str
+    previous_hash: str
+    record_hash: str
+
+    def __post_init__(self) -> None:
+        object.__setattr__(
+            self, "sequence", _non_negative_int(self.sequence, "sequence")
+        )
+        object.__setattr__(
+            self,
+            "model_digest",
+            _required_digest(self.model_digest, "model_digest"),
+        )
+        object.__setattr__(self, "suite_id", _identifier(self.suite_id, "suite_id"))
+        object.__setattr__(
+            self,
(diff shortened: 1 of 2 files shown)


DrVelvetFog (commented):
Read through this one. The chain construction and the metadata allowlist are both solid — but I think the "tamper-evident" claim is stronger than what the verifier can actually check today.

**The head isn't anchored to anything.**

`GENESIS_HASH` is a public constant derived from the schema version, so anyone who can write the ledger file can rebuild the whole chain and it verifies clean. I rebuilt a two-record ledger from scratch with better numbers:

```
honest ledger  verifies: True | scores: [0.71, 0.73]
FORGED ledger  verifies: True | scores: [0.94, 0.97]

$ python -m openmed.eval.provenance_ledger verify ledger.json
Verified 2 eval provenance record(s): ledger.json      # exit 0
```

That's the chain working as designed — record N commits to N−1 — but it only proves internal consistency. It catches a careless edit; it doesn't catch a rewrite. For eval results the realistic threat is someone with write access improving the numbers, which is the rewrite case.

The concrete gap is narrower than "this should be signed": `verify_ledger()` and the `verify` subcommand take no expected-head argument, so a caller who *does* have a trusted head hash has no way to check against it. `hmac` is imported, but only for `compare_digest`.

Two options, not exclusive:

- `verify_ledger(path, *, expected_head_hash=None)` plus `verify <ledger> --expected-head sha256:…`. Cheap, and it makes storing the anchor the caller's problem, which is the right call.
- Anchor it in the release. The `evidence` job in `publish.yml` already signs artifacts and attaches them to the tagged release — the ledger head is a natural thing to attach and sign there. Then verification doesn't have to trust the file it's checking.

Happy to do either.

**Smaller catch — the aggregate-suffix hatch swallows the sensitive-key rule.**

`_normalise_metadata_value` is strict in the right way: numbers, bools, null, nested, or a `sha256:` digest and nothing else, so raw text can't be a value. But in `_metadata_key`:

```python
if sensitive and not is_aggregate:
    raise UnsafeResultMetadataError(...)
```

any key ending in an `_AGGREGATE_SUFFIXES` token bypasses `_SENSITIVE_KEY_TOKENS` entirely:

```
rejected  'patient_name'
rejected  'mrn'
ACCEPTED  'patient_mrn_score'
ACCEPTED  'mrn_12345678_score'
ACCEPTED  'john_smith_score'
```

The intent reads as "allow `patient_count`, reject `patient_name`" — that's right, the hatch is just wider than the intent. Keying results by note or patient id is the kind of accident this guard exists to catch. Requiring each sensitive token to be *immediately followed* by the aggregate suffix would close the first two (`patient_count` ✓, `patient_mrn_score` ✗). `john_smith_score` is the reminder that a denylist can't catch a real name — if you want that closed, a registered metric-key schema is the only thing that does it.

maziyarpanahi:
You are right on both points — thanks for the concrete rewrite repro.

The current chain proves internal consistency, not authenticity of a ledger whose writer can recompute every hash. I will add `expected_head_hash` to `verify_ledger()` and `--expected-head` to the CLI in this PR, fail closed on malformed or mismatched values, and add a regression test where a fully rebuilt chain passes internal verification but fails against the previously trusted head. I will also narrow the module and CLI wording so the distinction between internal chain consistency and externally anchored verification is explicit. Attaching and signing that head in the release workflow can stay a follow-up because #886 explicitly puts cryptographic signing out of scope.

The aggregate-suffix bypass is also valid. I will remove the blanket exemption and add the cases you demonstrated. The immediate-follow rule closes `patient_mrn_score` and `mrn_12345678_score`, but you are right that it cannot make arbitrary keys PHI-safe: `john_smith_score` remains unknowable to a denylist. Rather than overclaim, I will make accepted metric keys schema-controlled or allowlisted for persisted result metadata and test that user-derived keys are rejected.

Thanks — these are both substantive fixes, not wording-only nits.

maziyarpanahi:
Both review points are resolved on the rebased head. Verification now accepts a separately trusted `expected_head_hash`, the CLI exposes `--expected-head`, malformed or mismatched anchors fail closed, and the regression suite proves that a coherently rebuilt chain passes internal verification but fails against its prior trusted head. I also removed the aggregate-suffix exemption in favor of an explicit v1 key schema; all three reported bypasses and arbitrary user-derived metric keys are rejected.

Validation is fully green: 20 focused tests, all 685 eval-package tests, 6,195 full-suite tests with 48 expected skips, canonical format/lint/type checks, scoped pre-commit, strict docs, package build, and the complete fresh hosted matrix across Linux, macOS, Windows, containers, security, and policy checks. OM-023 is confirmed complete via #338, and the final diff, metadata, and commit trailers are clean. Thanks for the precise rewrite repro and key examples — they materially improved the contract.

PR #1851, opened 2026-07-21
feat(multimodal): add RTF text extraction with char-offset spans (#856)

JonthanaHanh:
## Summary

Adds RTF (Rich Text Format) text extraction with char-offset source spans, enabling PII de-identification of RTF documents produced by legacy clinical and dictation systems.

Fixes #856

## What this PR does

- Adds `openmed/multimodal/documents_rtf.py` — a stdlib-only RTF parser (no third-party dependencies)
- Registers the `.rtf` handler in `openmed/multimodal/__init__.py`
- Adds `extract_rtf` to the package's `__all__` exports

## Implementation details

The parser handles:
- Control words (bold, italic, font size, etc.) — stripped from output
- Groups — nesting tracked, destination groups (fonttbl, colortbl, stylesheet, info, pict, etc.) skipped entirely
- Escaped characters (backslash, braces) — converted to literal
- Unicode escapes — decoded with signed 16-bit support
- Hex escapes — decoded
- Paragraph boundaries — converted to newlines
- Special characters (em-dash, en-dash, smart quotes, bullet) — decoded to Unicode

Each text fragment becomes a `SourceSpan` with metadata recording the byte range in the original RTF source, enabling span-level provenance for downstream redaction.

## Testing

- Parser tested with sample RTF containing paragraphs, special characters, and nested groups
- No third-party dependencies required (stdlib-only)
- Follows the same pattern as `documents_docx.py` and `epub.py`

PR #1774, opened 2026-07-19
feat: add ODT/ODF text extraction with offset maps

maziyarpanahi:
## Description

Adds dependency-free OpenDocument Text (`.odt`) extraction to the shared multimodal document contract. The extractor preserves paragraph and nested-list reading order, linearizes table cells with tabs and rows with newlines, and emits structural character-offset spans without retaining raw source text in metadata.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition/improvement

## Changes Made

- Added a standard-library ODT/ODF content-stream extractor with safe ZIP/XML validation and offset-aware handling for explicit ODF spaces, tabs, and line breaks.
- Registered `.odt` with `redact_document` and exported `extract_odt` from `openmed.multimodal` using the same `ExtractedDocument` surface as DOCX.
- Added synthetic coverage for reading order, nested lists, table linearization, offset round trips, registry discovery, contract parity, and malformed/unsafe inputs.
- Documented the public API, table behavior, supported scope, and unsupported ODS/ODP/write-back paths.

## Testing

- [x] I have added tests that prove the feature works
- [x] New and existing unit tests pass locally with my changes
- [x] `.venv/bin/python -m pytest tests/ -q` (`5205 passed, 46 skipped`)
- [x] `make format`
- [x] `make lint`
- [x] `make format-check`
- [x] `make type-check`
- [x] `make docs-build`
- [x] Scoped pre-commit hooks and repository policy checks

## Documentation

- [x] I have updated the documentation accordingly
- [x] I have added docstrings to new public functions
- [x] I have updated `CHANGELOG.md`

## Dependencies

- [x] I have not added any new dependencies

## Related Issues

Closes #857

## Screenshots/Examples

Not applicable; the new documentation includes a copy-paste extraction example.

Issue #857, opened 2026-06-28
Add ODT/ODF text extraction with char-offset map


maziyarpanahi:
## Summary
OpenDocument Text (.odt) is the native format of LibreOffice/OpenOffice and is common in non-Microsoft clinical and research settings. OpenMed extracts DOCX but not ODT, so ODF documents cannot be de-identified with offsets.

## Scope
- [ ] Add openmed/multimodal/odt.py extracting text from the ODF content stream with a char-offset map
- [ ] Preserve paragraph and list reading order; handle tables as linearized text
- [ ] Register the ingester in the multimodal ingest contract
- [ ] Mirror the DOCX extractor's contract for consistency

## Acceptance criteria
- [ ] Synthetic ODT extracts expected text in reading order with a correct offset map
- [ ] Tables linearize predictably and offsets round-trip
- [ ] Ingester is discoverable via the registry
- [ ] Contract matches the DOCX extractor surface

## Out of scope
- ODS spreadsheet or ODP presentation formats
- Re-rendering redacted ODT

## Files
- openmed/multimodal/odt.py
- tests/unit/multimodal/test_odt_extract.py
- docs/multimodal/odt-extraction.md

Parent epic: #1965
---
Task: OM-500  ·  Milestone: v2.2  ·  Priority: P3  ·  Size: S
Depends on: OM-064  ·  Blocks: —
Roadmap: v2.2 expanded trustworthy clinical data exchange inventory
Spec: PLANS/V2/EXECUTION/tasks/OM-500.md

PR #1853, opened 2026-07-22
feat(multimodal): add RTF text extraction with char-offset source spans

### LINKED ISSUE #856: Add RTF text extraction with char-offset map
## Summary
RTF is a common export format for legacy clinical and dictation systems, but OpenMed cannot extract plain text and offsets from it, so RTF documents cannot be de-identified with span provenance.

## Scope
- [ ] Add openmed/multimodal/rtf.py extracting plain text from RTF with a char-offset map back to source positions
- [ ] Handle control words, groups, and unicode escapes; strip formatting without losing text order
- [ ] Register the ingester in the multimodal ingest contract
- [ ] Round-trip offsets so redaction can locate spans in the original

## Acceptance criteria
- [ ] Synthetic RTF extracts to expected plain text with a correct offset map
- [ ] Control words and groups are handled without leaking markup into text
- [ ] Ingester is discoverable via the ingest registry
- [ ] Offset round-trip verified by a unit test

## Out of scope
- Re-rendering redacted RTF with original styling
- Embedded-image OCR (handled by OCR adapters)

## Files
- openmed/multimodal/rtf.py
- tests/unit/multimodal/test_rtf_extract.py
- docs/multimodal/rtf-extraction.md

Parent epic: #1965
---
Task: OM-499  ·  Milestone: v2.2  ·  Priority: P3  ·  Size: S
Depends on: OM-058, OM-064  ·  Blocks: —
Roadmap: v2.2 expanded trustworthy clinical data exchange inventory
Spec: PLANS/V2/EXECUTION/tasks/OM-499.md

kalra-mohit:
Closes #856.

RTF is still a common export format out of legacy clinical and dictation systems, and up to now `redact_document()` had no way to open one. This adds a `.rtf` handler that extracts plain text plus a byte-exact offset map back to the source file, so PII/PHI found in the extracted text can be traced back to the exact bytes that need redacting.

Scoped this to #856 only. #857 (ODT/ODF) already has an open PR (#1774) from @maziyarpanahi himself, so a second competing implementation of that didn't seem useful.

## Why stdlib, why latin-1

RTF's own control syntax (braces, control words) is plain ASCII; anything outside that goes through an escape — `\'hh` or `\uN` — never a raw multi-byte sequence. That means it doesn't need a binary parser the way DOCX or PDF do, so `openmed/multimodal/documents_rtf.py` is a plain stdlib scanner, no new dependency. It also means decoding the source file with `latin-1` (a lossless one-byte-to-one-character mapping) instead of UTF-8 gives every character offset the scanner produces for free as an exact byte offset into the original file, which is what a redaction write-back actually needs.

Destination groups — font/color tables, `\info`, headers/footers, embedded objects — get walked to keep brace nesting correct but never contribute to the extracted text. Skip state is a flag per group frame, inherited by child groups, rather than a single depth counter, so a destination nested inside another destination doesn't re-enable output the moment the inner one closes. `\*`, RTF's generic "ignorable destination" marker, is honored too, so a vendor-specific destination that isn't on the hardcoded keyword list still gets skipped instead of leaking its control words into the output.

## What I checked against the earlier attempt

There was a previous PR for this issue, #1851, that got closed for not meeting the quality bar. I went through that diff before starting mine, mostly so I wouldn't repeat the same mistakes, and it turned up several concrete bugs I made sure to handle here — each has its own test:

- Reading the file as UTF-8 instead of `latin-1` — corrupts or crashes on ordinary cp1252 content.
- Decoding `\'hh` hex escapes with `chr(byte_value)`, which is only correct for the low-ASCII range — cp1252 byte `0x92` is U+2019 (’), not U+0092. This version decodes through the document's declared code page (`\ansicpg`, defaulting to cp1252). (`test_cp1252_hex_escapes_decode_via_codepage_not_raw_codepoint`)
- `\uN` fallback handling only consumed a literal `?`; the far more common `\uN\'hh` pattern — a hex-escaped fallback byte, which is what Word actually emits — fell through and got appended as extra, duplicated text. (`test_unicode_escape_with_hex_fallback_is_not_duplicated`, `test_explicit_uc_controls_fallback_skip_width`)
- `\pard` mapped to a newline alongside `\par`. Since `\pard` conventionally shows up right after `\par` at the top of the next paragraph, that doubles every paragraph break in real documents — deliberately left out of the char-producing table here. (`test_par_immediately_followed_by_pard_does_not_double_newline`)
- A single skip-depth counter instead of a per-group flag, which can prematurely re-enable output for a destination nested inside another still-open destination. (`test_nested_skip_destination_does_not_leak_after_inner_group_closes`)
- No `\*` handling, so an unrecognized destination would leak raw control words as visible text. (`test_ignorable_star_destination_is_skipped_even_when_unrecognized`)
- No `\binN` handling — raw embedded binary got scanned as if it were RTF syntax, so a `{`, `}`, or `\` byte inside an OLE payload could desync brace nesting for the rest of the document. (`test_bin_binary_payload_is_consumed_verbatim`)

## The surrogate-pair bug

Once the first version was working, I threw some emoji-containing RTF at it to make sure non-Latin content survived round-trip, and hit a `UnicodeEncodeError` a few layers downstream. Traced it back to `\uN` handling: each `\uN` was being decoded on its own with `chr(codepoint & 0xFFFF)`, but astral-plane characters — anything above U+FFFF, which covers most emoji — can't be written as a single `\uN` in RTF. They're encoded as a UTF-16 surrogate pair: a high-surrogate `\uN` immediately followed by a low-surrogate `\uN`. Decoding each half independently produces two lone surrogate code points, which are invalid Unicode scalar values on their own and blow up the moment anything tries to UTF-8-encode the text later — exactly what a redaction write-back path does.

```mermaid
flowchart LR
    S["RTF source:<br/>\u-10179? \u-8704?<br/>(U+1F600 as a UTF-16 surrogate pair)"] --> Old & New

    subgraph Old["Old behavior"]
        O1["Decode \u-10179? alone<br/>-> chr(0xD83D)"] --> O2["Decode \u-8704? alone<br/>-> chr(0xDE00)"]
        O2 --> O3["Two lone surrogates emitted into text"]
        O3 --> O4["doc.text.encode('utf-8')<br/>raises UnicodeEncodeError"]
    end

    subgraph New["Fixed behavior"]
        N1["Decode \u-10179?<br/>-> 0xD83D, hold as pending high surrogate"] --> N2["Decode \u-8704?<br/>-> 0xDC00-0xDFFF: matching low surrogate"]
        N2 --> N3["Combine: 0x10000 + (high-0xD800)*0x400 + (low-0xDC00)<br/>= U+1F600"]
        N3 --> N4["Single real character emitted<br/>UTF-8 encode succeeds"]
    end
```

The fix holds a high surrogate when one is seen, instead of emitting it right away, and combines it with the next `\uN` escape if that one turns out to be the matching low surrogate — after letting each escape's own `\ucN` fallback run get consumed, since those fallback bytes sit between the two `\uN` tokens in the source. An unpaired high surrogate or an orphan low surrogate is malformed or truncated input, and gets replaced with U+FFFD rather than emitted raw, matching how the scanner already handles undecodable `\'hh` bytes.

Added tests for the paired case (U+1F600, grinning face, via `\u-10179?\u-8704?`) plus the unpaired-high and orphan-low malformed cases, and updated `docs/multimodal/rtf-extraction.md` to describe the behavior.

## Testing

`tests/unit/multimodal/test_rtf_extract.py` — 21 tests against hand-built synthetic RTF (no real PII): the regressions above, body-text reading order, metadata/table exclusion, a hyperlink field (`\fldinst` skipped, `\fldrslt` kept), table-row linearization (`\cell` -> tab, `\row` -> newline), byte-exact offset round-trips (locate a substring in the extracted text, then check its span's `source_start`/`source_end` against the literal bytes at that position — including through a hex-escaped curly quote), the surrogate-pair cases above, an empty `{\rtf1}` document, malformed non-RTF input rejection, and dispatch through `redact_document()`.

Also ran the full suite (`.venv/bin/python -m pytest tests/ -q` → 5972 passed, 59 skipped, 0 failed), `ruff check .` / `ruff format --check .`, scoped `mypy`, and `mkdocs build --strict` — all clean.

## One thing worth flagging

Hex-escape decoding batches up consecutive `\'hh` runs and decodes them together through the active code page, which matters for a genuinely multi-byte code page (e.g. Shift-JIS, `\ansicpg932`) but I've only actually exercised the single-byte case (cp1252) here, since that covers the overwhelming majority of real clinical/dictation RTF. I don't have a realistic multi-byte-codepage RTF sample to validate against, so I'd treat that path as implemented but not independently proven yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added dependency-free RTF (Rich Text Format) extraction with a stdlib-only
+  scanner that decodes `\'hh` hex escapes by document code page instead of
+  raw code point, consumes `\uN` Unicode-escape fallback runs instead of
+  duplicating them, honors `\*` ignorable destinations and `\binN` raw binary
+  payloads so embedded objects cannot desync group nesting, and emits
+  byte-exact character-offset source spans for redaction (#856).
 - Added build-generated `llms.txt` and `llms-full.txt` documentation feeds with
   curated quickstart, API, de-identification, agent, MCP, and REST coverage,
   plus strict local and Pages build checks (#1787).
--- a/docs/multimodal/rtf-extraction.md
+++ b/docs/multimodal/rtf-extraction.md
@@ -1,0 +1,64 @@
+# RTF Extraction
+
+OpenMed can extract text from RTF (Rich Text Format) documents through the
+shared multimodal document contract. RTF is a common export format for legacy
+clinical and dictation systems. The ingester is stdlib-only: RTF's own syntax
+is plain ASCII, so no third-party parser is required to read it.
+
+```python
+from openmed.multimodal import extract_rtf
+
+document = extract_rtf("dictation-note.rtf")
+
+print(document.text)
+```
+
+Each `SourceSpan` maps a range in `document.text` back to the exact byte
+range in the original RTF file:
+
+```python
+offset = document.text.index("John Q. Public")
+span = document.location_at(offset)
+
+print(span.metadata["source_start"], span.metadata["source_end"])
+```
+
+The metadata is PHI-safe: spans carry numeric byte offsets only, never a copy
+of the raw source text.
+
+## Behavior
+
+- `.rtf` files are discoverable through `redact_document`.
+- Font, color, style, and list tables, document `\info` metadata (author,
+  title, ...), headers, footers, footnotes, annotations, and embedded
+  picture/object payloads are walked (to keep group nesting correct) but
+  never contribute to extracted text.
+- `\uN` Unicode escapes are decoded to the real character; the
+  `\ucN`-controlled fallback run that follows (for older readers) is
+  consumed rather than duplicated into the output.
+- `\'hh` hex escapes are decoded using the document's declared code page
+  (`\ansicpg`, defaulting to Windows-1252) rather than treated as raw
+  Unicode code points, so escapes like curly quotes and em dashes decode to
+  the correct character.
+- `\*` marks a destination as ignorable; any such group is skipped even when
+  its control word is not one OpenMed recognizes by name.
+- `\binN` raw binary payloads (used by embedded OLE objects) are consumed
+  verbatim, so arbitrary bytes inside them cannot be misread as RTF syntax
+  and desynchronize group nesting for the rest of the document.
+- `\par` maps to a newline; `\pard` (paragraph formatting reset, which
+  commonly appears immediately after `\par`) does not, so paragraphs are not
+  double-spaced.
+- Table cells are linearized with `\t` per cell and `\n` per row, matching
+  how DOCX and ODT table extraction are documented.
+
+The source file is decoded with `latin-1`, not UTF-8: RTF's control syntax is
+always ASCII, and anything outside that range is represented through an
+escape, never a raw multi-byte sequence. `latin-1` is therefore a lossless,
+one-byte-to-one-character decode, which is what lets every span's
+`source_start`/`source_end` be an exact byte offset into the original file
+rather than an offset into some intermediate re-encoding.
+
+Re-rendering a redacted RTF file with its original styling, and OCR of
+embedded images, are out of scope; callers can use source offsets to locate
+and redact the corresponding bytes directly when they need custom write-back
+behavior.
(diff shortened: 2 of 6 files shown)


kalra-mohit:
@maziyarpanahi whenever you get a chance to take a look, happy to address any feedback!

maziyarpanahi:
Hi @kalra-mohit 
Thank you for your contributions. Could you please re open the PR and make sure you don't have claude as co-author? We have a no coding agent names in the title, name of the branches, nor as a co-author. Everyone is using coding agents which is totally fine and even encouraged, but only one of them include itself as a co-author which is not fair to other agents. 

Thanks again, that would also fix the conflict issue

PR #1912, opened 2026-07-30
feat(grounding): language-aware sparse candidate generation with source-language provenance

PouyanJay:
> **Stacked on #1911** (`feat/grounding-alias-candidate-generator`, issue #1796). Until #1911 merges, this PR's diff shows #1796's commit as well as this one; please merge #1911 first, after which this reduces to the single language-aware commit. The review below concerns only that incremental commit.

## Summary

Makes the sparse alias candidate generator (from #1796) **language-aware**, threading source-language provenance end-to-end so a non-English mention is matched against language-appropriate aliases and every emitted `Candidate` carries the correct `source_language`.

- `SparseCandidateGenerator.generate` / `generate_candidates` gain a `language` kwarg. It resolves `source_language = normalize_language(language)`, drives the exact stage via `index.lookup_all(query, language=source_language)`, builds the fuzzy n-gram index from the language's aliases, maps fuzzy hits back with the same language, and stamps `source_language` on every candidate (the field already exists in the #1796 `Candidate` contract — no new field).
- English default (`None`/`"en"`) is **byte-for-byte unchanged** (verified cross-branch, see below).

Closes #1799.

## Scope note

The cross-lingual alias space itself — language-tagged alias columns, per-language indexing, `lookup_all(language=)`/`aliases_for_language(...)` with English fallback, the synthetic `grounding_crosslingual.jsonl` fixture, and the linker path — already shipped in #1292. #1799 is therefore an **extension**: the only untouched gap was #1796's sparse generator. Per the issue's additive charter this PR touches just `candidate_generator.py` + the crosslingual test; `vocab.py`/`registry.py`/`linkers/base.py` were already language-aware and are left unchanged.

## Post-review hardening (independent review pass)

- **Fuzzy stage now keeps English-only concepts reachable under a non-English language.** Previously the fuzzy n-gram vocabulary for e.g. `es` contained only `es` aliases, so a misspelled mention of an English-only concept was unreachable under `es` (the exact stage already fell back to English per-concept; the fuzzy stage did not). The fuzzy alias source now unions the language aliases with the English aliases. Added `test_fuzzy_stage_reaches_english_only_concept_under_non_english_language` (mixed-coverage index).

## Known limitations / follow-ups (disclosed)

[...]

PR #1911, opened 2026-07-30
feat(grounding): sparse alias candidate generator with provenance-carrying Candidate contract

PouyanJay:
## Summary

Adds a deterministic, offline sparse alias candidate generator over the free clinical vocabularies, and extends the grounding `Candidate` type with provenance so downstream retrieval/ranking can trace and rank candidates. This is the foundation the rest of the grounding roadmap builds on.

- **`openmed/clinical/grounding/types.py`** — extends `Candidate` **additively** with `source=""`, `matched_alias=None`, `match_kind=None`, `vocab_version=None` (all defaulted, appended after the existing five fields) and a read-only `concept_id` property returning `code`. Positional construction and existing consumers are unchanged.
- **`openmed/clinical/grounding/candidate_generator.py`** — `SparseCandidateGenerator` + `generate_candidates(mention, systems, k=10, *, loader=None, ngram_size=3, min_similarity=0.3) -> list[Candidate]`. Unions exact normalized-alias hits (score 1.0) with character-trigram TF-IDF fuzzy hits, dedups per `(system, code)` keeping the best score, and orders by `(-score, system_priority_rank, code)`. Pure-Python (no `rapidfuzz`) for byte-identical determinism. Registered via `register_linker("sparse", ...)`.
- **`openmed/clinical/grounding/vocab.py`** — recognizes `concept_id`/`canonical_term` alias-table keys; adds `VocabularyIndex.content_hash -> "sha256:<hex>"` used as the `vocab_version` provenance key.
- Grounding `__init__` exports; a synthetic 45-concept alias fixture; unit tests.

Closes #1796.

## Deviations / judgment calls

- **Kept `code`; did not rename to `concept_id`.** The merged `Candidate` already uses `code` and is imported by 4 linkers, the FHIR exporter, `medication_sig`, and `icd11_api`; renaming would be a breaking change. Added a `concept_id` read-only property satisfying the schema wording without breakage.
- **Provenance fields are additive with defaults**, so all existing `Candidate(...)` sites and equality/determinism tests are untouched.
- **Did not edit `registry.py`.** Every existing linker self-registers by calling `register_linker` from its own module; the generator follows that idiom, and `get_linker("sparse")` resolves it (no name collision with `hpo`/`icd10cm`/`rxnorm`).
- **Fuzzy matching is self-authored char-trigram TF-IDF**, not the optional `rapidfuzz` path — guarantees cross-run byte-identity without the optional `grounding` extra.
[...]

Issue #1799, opened 2026-07-20
Add a language-aware alias space and source-language provenance to grounding


maziyarpanahi:
## Summary
Advances OM-751 by making the free-vocab loader and candidate generator language-aware so a Spanish 'diabetes mellitus tipo 2' or French 'infarctus du myocarde' resolves to the same canonical code as its English equivalent. Today aliases are English-only, so non-English NER cannot ground to shared international codes. This delivers the lexical cross-lingual bridge (no embeddings) plus a source-language field on every grounded record.

## Scope
- [ ] Extend openmed/clinical/grounding/vocab.py so alias rows carry a language column keyed to the same canonical code; bundle only English + synthetic non-English aliases, treat user-supplied UMLS/SNOMED multilingual aliases as out-of-process.
- [ ] Thread a language code from the pipeline/router through registry.py and linkers/base.py so candidate generation selects the target-language alias space.
- [ ] Make the sparse generator score against the target-language aliases first, falling back to the English alias set (lexical only) when a target-language alias is absent.
- [ ] Emit source_language on each grounded record and preserve existing English grounding outputs byte-for-byte where aliases are unchanged.
- [ ] Commit synthetic cross-lingual gold (non-English mention -> canonical code) with only synthetic mentions and permissive aliases, plus a leakage check the alias tables embed no restricted corpus text.

## Acceptance criteria
- [ ] A non-English mention in es/fr/de/zh resolves to the same canonical ICD-10-CM/RxNorm/HPO code as its English equivalent on the synthetic cross-lingual gold.
- [ ] Per-language top-1 grounding accuracy on the synthetic gold is >= 0.80 for each of es/fr/de/zh with the English baseline non-regressed.
- [ ] The default install performs lexical cross-lingual grounding with no network access; every grounded record carries a source_language field (unit-tested).
- [ ] A leakage-first test confirms no restricted/DUA corpus text is present in the committed alias tables (synthetic + permissive only).
- [ ] test suite green: .venv/bin/python -m pytest tests/unit/clinical/test_grounding_crosslingual.py -q

## Out of scope
- The opt-in multilingual embedding fallback for zero-alias mentions (deferred; graceful lexical-only default here).
- Non-Latin script normalization beyond the committed synthetic set.
- Rerank changes (owned by the reranker task).

## Files
- openmed/clinical/grounding/vocab.py
- openmed/clinical/grounding/registry.py
- openmed/clinical/grounding/linkers/base.py
[...]

Issue #1796, opened 2026-07-20
Add a free-vocabulary alias candidate generator with a provenance-carrying Candidate contract


maziyarpanahi:
## Summary
Advances the OM-040 grounding epic by delivering stage (a)+(b): the sparse candidate generator every downstream ranker and exporter depends on. Today grounding/types.py defines a Candidate but there is no deterministic generator that turns a mention into scored candidates over a user-loaded free vocabulary (RxNorm, ICD-10-CM, LOINC, HPO), and no bundled UMLS/SNOMED is permitted. This is the first mergeable piece: an offline alias-index + TF-IDF matcher that emits Candidate objects with source and score.

## Scope
- [ ] Extend openmed/clinical/grounding/vocab.py with a loader that reads user-supplied free-vocab alias tables (concept_id, system, canonical_term, aliases) and builds an in-memory normalized alias index; ship no vocabulary content in the wheel.
- [ ] Add generate_candidates(mention, systems, k) in openmed/clinical/grounding/candidate_generator.py that unions exact-alias hits and character-ngram TF-IDF fuzzy hits into ordered Candidate(concept_id, system, score, source='sparse') objects.
- [ ] Populate provenance fields on each Candidate: matched_alias, match_kind (exact|fuzzy), and vocab_version content-hash so downstream can audit the source.
- [ ] Add deterministic tie-break (score desc, then system priority, then concept_id) so identical inputs always yield identical ordering.
- [ ] Register the generator through grounding/registry.py so linkers resolve it via get_linker without importing it directly.

## Acceptance criteria
- [ ] generate_candidates('type 2 diabetes', systems=['icd10cm']) returns Candidate objects with the correct E11 family concept in the top result on a synthetic gold vocab.
- [ ] On a 200-mention synthetic entity->code gold, top-5 recall of the sparse generator is >= 0.85 and results are byte-identical across two runs (determinism).
- [ ] Every returned Candidate carries source='sparse', matched_alias, match_kind, and a non-empty vocab_version hash (unit-tested).
- [ ] A unit test asserts the loader raises a clear error on an unknown system and that no vocabulary rows are importable from package data (wheel stays content-free).
- [ ] test suite green: .venv/bin/python -m pytest tests/unit/clinical/grounding/ -q

## Out of scope
- Dense/embedding retrieval and ANN indexing (separate sibling task).
- Cross-encoder reranking and section-aware scoring.
[...]

PR #1917, opened 2026-07-30
feat(normalization): opt-in multilingual embedding fallback for grounding (OM-751b)

### LINKED ISSUE #1289: OM-751b: Add opt-in multilingual embedding fallback for grounding
Parent epic: #1247

## Summary
Add the optional zero-alias fallback slice for OM-751: an offline multilingual clinical embedding backend that can be enabled explicitly while keeping the default install lexical and dependency-light.

## Scope
- Add an optional backend in openmed/clinical/normalization/backend.py for multilingual similarity when no target-language alias exists.
- Keep embedding dependencies optional and disabled by default.
- Make disabled or unavailable embedding backends degrade gracefully to lexical-only ranking.
- Add tests for opt-in behavior, unavailable backend handling, and offline default behavior.

## Acceptance criteria
- Default install performs grounding without importing or requiring the embedding backend.
- Enabling the fallback uses a local/offline backend only and never performs network access.
- Missing optional dependencies produce a clear fallback result rather than an exception during normal lexical grounding.
- Tests cover disabled, unavailable, and enabled fallback paths with synthetic inputs.

Task: OM-751b

PouyanJay:
> **Stacked on #1912 → #1911.** This branch sits on #1799 (#1912) → #1796 (#1911). Until those merge, this PR's diff also shows their commits. Merge order: **#1911 → #1912 → this**.

## Summary

Adds an **opt-in** multilingual embedding fallback for grounding — an offline embedding backend that can be enabled explicitly while the **default install stays lexical and dependency-light**. It kicks in only as a zero-alias fallback (when no target-language alias exists).

- **`openmed/clinical/normalization/backend.py`** —
  - `MultilingualEmbeddingBackend(enabled=False, embedder=None, model_path=None, dependency="sentence_transformers")` — **disabled by default**; the optional lib is resolved **lazily** (only inside `status()`/`_load_embedder`, never at module import); an injected `embedder` (any `MultilingualEmbedder` Protocol impl) bypasses the import entirely.
  - `status() -> EmbeddingBackendStatus` never raises (reasons: disabled / not-installed / no-model / ready).
  - `rank(mention, concepts) -> EmbeddingFallbackResult` — pure-Python cosine re-rank; `used_embedding=False` (concepts unchanged) whenever disabled/unavailable/failed, so it never raises during normal grounding.
  - `fallback(mention, concepts, *, lexical_candidates=())` — returns the lexical candidates unchanged when present, and only consults embeddings for the empty (zero-alias) case.
- Additive `normalization/__init__.py` exports; a new offline/deterministic test suite.
- **No `pyproject.toml` / `uv.lock` change** — the design adds no declared dependency, so the lockfile gate is not engaged (deliberately avoiding a drift like a sibling PR hit).

Closes #1289.

## Post-review hardening (independent review pass)

The review confirmed the critical guarantee — a default `import openmed` + lexical grounding imports **no** heavy embedding lib (`sentence_transformers`/`torch`/`numpy` absent from `sys.modules`) — and that every degradation path returns cleanly without raising. Two hardening nits fixed:

- **[MEDIUM] Offline-only is now enforced in code, not just documented.** `_load_embedder` now gates on `os.path.exists(model_path)` **before** importing the dependency, so a non-local path (e.g. a remote model id) is treated as unavailable and can never trigger a download. Added `test_remote_model_path_is_not_loaded_or_downloaded`.
- **[LOW] Ragged embedder output no longer silently drops candidates.** `rank()` now asserts `len(vectors) == len(pool) + 1` before zipping; a mismatch degrades to lexical instead of returning a truncated pool. Added `test_ragged_embedder_output_degrades_to_lexical_without_dropping`.

## Deviations / judgment calls

- **No `pyproject` extra added.** The embedding lib is gated purely by lazy import + an injectable embedder, satisfying every acceptance criterion while keeping the lock untouched. The dependency name is a constructor arg so a future extra can be wired without an API change.
- **Operates on `TerminologyConcept`** (the normalization module's type named by the issue), not the grounding `Candidate` — kept self-contained in `normalization/backend.py`.
- **Pure-Python cosine** (no numpy), mirroring the repo's dependency-free lexical matcher, for deterministic offline scoring.

## Data provenance

All test data is synthetic (algorithmic stub embedders, `SYNTHETIC_CONCEPTS`). No real PHI, no bundled terminology, no model downloads.

## Verification

- **Import-isolation (the key guarantee): PASS** — verified no optional embedding lib is imported at module load or during a default lexical fallback.
- `make test`: **8878 passed, 82 skipped, 27 failed** — all 27 the documented pre-existing env-only failures. The embedding-fallback suite: 9 passed. `test_normalization_repository_scan` + `test_public_api_docstrings` gates green; no new mypy errors.
- `pre-commit` on changed files: all hooks pass. `pyproject.toml`/`uv.lock` untouched. `git rev-list --count feat/grounding-language-aware-alias..HEAD` = 1.

## Relationship

Wave 3 (OM-751b) of the Concept-Grounding roadmap, stacked on the language-aware alias work (#1799/#1912). Exposes the zero-alias fallback as a library primitive; wiring it into the higher-level pipeline is OM-751c (#1290), and cross-lingual eval metrics are OM-751d (#1291).


maziyarpanahi:
Thanks for the careful offline-fallback hardening here. I rechecked this exact head against current master: child issue #1289 is in v2.1, the merge simulation is clean, local-only model-path enforcement and ragged-output degradation are covered, every hosted check is green, and there are no unresolved review threads. This is ready to merge.

PR #1074, opened 2026-06-29
Add deterministic content-addressed sharding for batch de-identification

maziyarpanahi:
# Pull Request

## Description
Adds the first OM-632 child: deterministic content-addressed shard planning for distributed batch de-identification. The new API assigns documents to stable shards from a stable id hash, keeps worker count out of membership decisions, rejects duplicate or missing ids before processing, and emits PHI-minimized plan metadata for later manifest/report work.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition/improvement

## Changes Made
- Added `openmed.processing.distributed` with stable document hashing, shard assignment, shard plans, shard metadata, and safe sharding errors.
- Exported the sharding API from `openmed.processing`.
- Added unit tests for input-order stability, fixed-shard worker-count independence, duplicate and missing id rejection, object/custom id extraction, and PHI-free emitted metadata.

## Testing
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this change with different models/inputs

Commands run:

- `make format`
- `make lint`
- `make format-check`
- `.venv/bin/python -m pytest tests/unit/processing/test_distributed_sharding.py tests/unit/test_batch.py tests/unit/processing/test_batch_progress_callback.py -q`
- `.venv/bin/python -m pytest tests/ -q`

## Documentation
- [ ] I have updated the documentation accordingly
- [x] I have added docstrings to new functions/classes
- [ ] I have updated the CHANGELOG.md

## Code Quality
- [x] I ran `make format`, `make lint`, and `make format-check`
- [ ] For Swift/OpenMedKit changes, I ran `make format-swift` and `make lint-swift`
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings

## Dependencies
- [x] I have not added any new dependencies
- [ ] OR I have added new dependencies and they are justified because: ____

## Checklist
- [x] I have read the contributing guidelines
- [x] My commits have clear, descriptive messages
- [x] I have squashed/organized my commits appropriately

## Related Issues
Closes #1068
Part of #995

## Screenshots/Examples
Not applicable.

Issue #1073, opened 2026-06-29
Add PHI-free reports and CLI resume path for distributed batch runs


maziyarpanahi:
Parent: #995

## Summary
Add operator-facing reporting and CLI/programmatic resume controls for distributed batch de-identification runs.

## Scope
- Emit aggregate per-shard counts, durations, failures, and run fingerprint.
- Add a CLI path to start or resume a distributed run by run id.
- Add scanner coverage proving no document text appears in manifests, reports, or logs.
- Document the local-first operational workflow.

## Acceptance criteria
- Reports contain counts, durations, failures, fingerprints, and no raw document text.
- Operators can resume a run from the CLI and programmatic API.
- Scanner tests fail if document text is written to manifest or report artifacts.

Part of #995.

Issue #1072, opened 2026-06-29
Add Ray and Spark adapters for distributed batch de-identification


maziyarpanahi:
Parent: #995

## Summary
Add optional Ray and Spark executor adapters behind the distributed batch Executor protocol.

## Scope
- Lazily import Ray and Spark dependencies only when their adapters are used.
- Run the same sharding and resumability conformance tests across local, Ray, and Spark backends.
- Keep adapter failures isolated from the core local-first package when optional dependencies are absent.

## Acceptance criteria
- Importing openmed.processing does not require Ray or Spark.
- Ray and Spark adapters pass the distributed batch conformance suite when dependencies are installed.
- Adapter errors are actionable and do not leak document text.

Part of #995.

Issue #1071, opened 2026-06-29
Add local executor and idempotent shard writes for distributed batch runs


maziyarpanahi:
Parent: #995

## Summary
Add the executor abstraction and local multiprocessing backend for running shard work with at-least-once execution and idempotent writes.

## Scope
- Define the Executor protocol used by distributed batch processing.
- Implement a local multiprocessing executor.
- Write each shard to a temporary output path and atomically rename on success.
- Reduce per-shard summaries into a run-level result.

## Acceptance criteria
- Local execution produces the same per-document output digests as single-process processing.
- Re-executing a shard overwrites the same final output rather than appending.
- Partial temporary outputs are not counted as completed shard outputs.
- Tests cover worker failures during shard execution.

Part of #995.

PR #2027, opened 2026-08-02
feat: add a durable distributed batch run manifest

### LINKED ISSUE #1069: Add a durable distributed batch run manifest
Parent: #995

## Summary
Add a durable manifest for distributed batch runs that records shard state and output digests without storing document text.

## Scope
- Track shard status, attempts, timing, output paths, and output digests.
- Persist manifest updates durably with atomic writes.
- Skip completed shards on resume after validating their recorded outputs.
- Detect missing or corrupted shard outputs before completing a run.

## Acceptance criteria
- A manifest written after a partial run can be loaded and validated.
- Completed shards are skipped on resume when their output digest still matches.
- Missing or mismatched output digests mark shards for re-execution.
- Manifest JSON contains no document text.

Part of #995.

PouyanJay:
Closes #1069

Part of #995 (OM-632). Adds a durable manifest for distributed batch runs that records shard state and output digests without storing document text.

## Summary

- **Tracks shard status, attempts, timing, output paths and output digests.** `ShardRecord` carries `shard_id`, `fingerprint`, `document_count`, `status`, `attempts`, `started_at`/`completed_at`, `output_path`, `output_digest`, `output_bytes`, `worker_id` and `error_type`.
- **Persists manifest updates durably with atomic writes.** `LocalFileRunManifestStore` reuses the existing `openmed/processing/batch.py::_atomic_write_bytes` sequence — same-directory temp file, flush, `fsync`, `os.replace`, then directory `fsync` — so a reader always observes either the previous manifest or the new one.
- **Skips completed shards on resume after validating their recorded outputs.** `shards_to_execute()` returns a completed shard to the work queue only when its recorded digest no longer matches.
- **Detects missing or corrupted shard outputs before completing a run.** `validate_shard_outputs()` reports `valid` / `missing` / `mismatched`, and with `strict=True` raises `ShardOutputMissingError` / `ShardOutputDigestMismatchError`.

This module is a **record and validator only**. It never executes shards and never increments `attempts`; executor wiring belongs to the follow-up children.

## Public surface

Eighteen names, all re-exported from `openmed.processing`:

`RUN_MANIFEST_SCHEMA_VERSION` (int `1`, mirroring `checkpoint.py::CHECKPOINT_SCHEMA_VERSION` and `batch.py::_BATCH_CHECKPOINT_SCHEMA_VERSION`), `MAX_LABEL_LENGTH`, `ShardStatus`, `ShardRecord`, `BatchRunManifest`, `ShardOutputValidation`, `RunManifestStore` (runtime-checkable `Protocol`), `InMemoryRunManifestStore`, `LocalFileRunManifestStore`, `RunManifestError`, `ManifestSchemaError`, `UnknownShardError`, `ShardOutputMissingError`, `ShardOutputDigestMismatchError`, `build_run_manifest`, `shard_output_digest`, `validate_shard_outputs`, `shards_to_execute`.

Output digests use the repo's existing `openmed/core/model_integrity.py::sha256_file`, so they carry the canonical `sha256:<hex>` prefix rather than a bare hex string.

## No document text or identifiers

Every field is an integer, a timestamp, an operator-chosen relative path, a version string, or a SHA-256 digest. Deliberately **excluded**: document text, `document_ids`, `document_hashes` (which are id-derived), and exception **messages** — only `error_type`, the exception class name, is stored, because messages routinely echo record content.

Four construction-time invariants enforce this rather than leaving it to convention:

1. `error_type` must be a bounded dotted identifier — passing `str(exc)` where `type(exc).__name__` was meant is a hard `RunManifestError`.
2. `output_path` must be relative, must not contain `..`, and must not contain control characters.
3. `status=COMPLETED` requires both an `output_path` and an `output_digest`.
4. `output_digest` must carry the `sha256:` prefix.

`run_id` and `worker_id` are operator-supplied free text written verbatim, so both are bounded at `MAX_LABEL_LENGTH` (128) and reject control characters.

The `error_type` guard is documented honestly as a **syntactic shape check, not a content guarantee** — `test_error_type_guard_is_a_shape_check_not_a_content_guarantee` asserts that `Patient_Jane_Doe_MRN_12345` **is** accepted, alongside what the guard does reliably stop. A test that pins the guard's limitation is more useful than one that overstates it.

### 1. Deserialization was a laundering vector

The first draft of `from_dict` type-checked integers and floats strictly but passed every **string** field through a bare `str()`, which never fails. That was a PHI-laundering path, demonstrated end-to-end through the real store:

```
worker_id = {"note": "<record-derived text>"}  -> rendered to a string and re-saved to disk
run_id    = None                              -> the literal "None"
fingerprint = None                            -> "None", which is non-empty and so passed the guard
output_path = ["a", "b"]                      -> "['a', 'b']", which then passed path validation
```

The `fingerprint` case was the worst: a null-fingerprint manifest round-tripped looking valid, with the same fake fingerprint on every shard, defeating the plan binding that resume correctness depends on.

Both classes now **reject rather than coerce**, at the boundary where foreign bytes enter:

```
CLOSED worker_id dict:      ManifestSchemaError: worker_id must be a string
CLOSED run_id None:         ManifestSchemaError: run_id must be a string
CLOSED fingerprint None:    ManifestSchemaError: fingerprint must be a string
CLOSED output_path list:    ManifestSchemaError: output_path must be a string
```

Relatedly, both regexes originally used `.match()` with `$`, which matches before a trailing newline. `sha256:<64hex>\n` was accepted at construction, round-tripped through JSON, and then never equalled `sha256_file()` output — so the shard was classified `mismatched` **forever**, requeued on every resume, with no error explaining why. `error_type="RuntimeError\n"` survived to disk as a line-injection primitive. Both now use `re.fullmatch`, and control characters are rejected in every free-text field.

### 2. Do not unify these fields onto the input gateway

`openmed/utils/gateway.py` (added in #2023) centralizes input normalization, and routing manifest fields through `gateway.normalize_text()` looks like an obvious later cleanup. **It would re-introduce the bug above**: `normalize_text` funnels through `_coerce_to_text`, whose `str(value)` at `gateway.py:141` is exactly the coercion that renders a nested dict into a stored string.

The reject-don't-coerce stance here is not in tension with the gateway — it matches the gateway's own `validate_language` (`gateway.py:276`), which requires `isinstance(lang, str)` and refuses to coerce. The gateway's split is principled: coercion for the **document text payload**, strict typing for **identifier-class input**. Every manifest field is identifier-class. The gateway's PHI stance (`gateway.py:5-6`, errors never echo input text) is the same reason `error_type` stores a type name and not a message.

## Judgment calls

- **`MAX_LABEL_LENGTH = 128` versus the gateway's `DEFAULT_MAX_TEXT_CHARS = 1_000_000`.** Different units and different concerns — a manifest label versus a document payload — so the numbers are deliberately unrelated rather than inconsistent.
- **This bound is deliberately not environment-overridable**, unlike the gateway's limits (`OPENMED_MAX_TEXT_BYTES` and friends). A manifest label bound is an internal invariant; making it tunable would let an operator widen the very injection surface that PHI-free run reporting depends on.
- **`LocalFileRunManifestStore` exposes no `fsync` toggle.** `_atomic_write_bytes` fsyncs unconditionally, so a `fsync=False` parameter would either be inert or require duplicating a weaker write path. It takes an `atomic_write_hook` instead, which is what makes the crash-point tests possible.
- **`UnknownShardError` derives from both `RunManifestError` and `KeyError`.** It is raised where a plain `KeyError` used to be, so callers written against either hierarchy keep working; an explicit `__str__` suppresses `KeyError`'s quoted representation.
- **`shards_to_execute()` returns `RUNNING` shards**, and this is documented in the docstring rather than left implicit. Deciding whether a `RUNNING` shard is a straggler still making progress or an orphan left by a crashed worker needs liveness information this module does not have; that policy belongs to the executor. Empty shards are likewise returned and can be filtered with the existing `ShardPlan.non_empty_shards`.

## Deviations from the issue

The issue specifies scope and acceptance criteria but does not list files, so the complete file list is given here with the reason each is required.

| File | Why |
|---|---|
| `openmed/processing/run_manifest.py` (new) | The module. A separate file rather than an extension of `distributed.py`, because the follow-up children stack on this branch and a separate module minimizes conflicts; `checkpoint.py` sets the precedent of one durable-state module per concern. |
| `openmed/processing/__init__.py` | Re-exports. Forced by the package's own convention — the closest merged analog, deterministic batch sharding (#1074, `9c62a9ed`), added 20 exports here in the same way. |
| `tests/unit/processing/test_batch_run_manifest.py` (new) | Proving tests, 51 of them. |

Three files total, matching #1074's restraint exactly (`openmed/processing/distributed.py`, `openmed/processing/__init__.py`, `tests/unit/processing/test_distributed_sharding.py`).

**Auto-enrolling gates checked that did *not* force additional files**, each verified by execution rather than assumed:

- `scripts/release/public_api_baseline.json` — **not** regenerated. `scripts/release/check_public_api.py` classifies new symbols as `kind="added"` (lines 482, 499, 561) and only unannounced **breaks** fail; `tests/unit/release/test_public_api_compat.py::test_current_surface_matches_committed_baseline` asserts `failures == []`. #1074 likewise added 20 exports without touching the baseline. Verified passing.
- `tests/unit/release/test_api_surface_diff.py` — passes unchanged.
- `tests/unit/release/test_import_budget.py` — no new cost; `openmed.core` is already imported through `openmed/processing/advanced_ner.py:12`. Verified passing.
- `tests/unit/test_no_telemetry.py` — the module is standard-library-only plus two in-repo imports, so it does not enter the reviewed network-surface allowlist. Verified passing.
- **No documentation page added.** `docs/api-reference.md` does not enumerate these symbols, so there is no drift gate to satisfy; conversely, adding any new `docs/*.md` **would** trip `tests/unit/test_docs_publication.py::test_publication_classifies_every_default_markdown_page_exactly_once`, which requires every default markdown page to be classified exactly once in the nav.
- **No CLI surface**, so `tests/unit/interop/test_tool_schema_sync.py` and the `--json` envelope contract are untouched. CLI resume belongs to a later child.

## Windows portability

CI surfaced one genuine portability bug in this branch's own tests: the unreadable-manifest case used `chmod(0o000)`, which does **not** make a file unreadable to its owner on Windows — the POSIX permission bits are not enforced for owner reads, so `open()` succeeded, no `OSError` was raised, and the `RunManifestError` assertion never fired. Root bypasses the bits the same way on POSIX.

Rather than delete the case or weaken the assertion, it is split so that the **product contract stays verified on every platform** and only the platform-specific *mechanism* is skipped:

- `test_read_failures_surface_as_run_manifest_error_on_every_platform` injects the `OSError` directly (parametrized over `PermissionError` and a generic `OSError`). This is what the test is actually for — that no read failure escapes the `RunManifestError` hierarchy, so a caller recovering with a single `except RunManifestError` cannot crash. It runs everywhere.
- `test_posix_permission_bits_make_a_manifest_unreadable` keeps the real `chmod` mechanism, skipped with a stated reason where the bits are not enforced (Windows, or running as root).
- The invalid-UTF-8 and directory cases were already portable and continue to run on all platforms — opening a directory raises `OSError` on Windows too, which is why that assertion passed there.

The assertion text is unchanged in every case; nothing was loosened to make CI green.

## Verification

Full suite, complete output captured and the sorted `FAILED` list diffed against the canonical baseline measured at `2965ec10`:

|---|---|---|---|---|
| baseline at `2965ec10` | 31 | 9816 | 86 | 1 |
| this branch | 31 | 9867 | 86 | 1 |

The two `FAILED` sets are **identical** — no new failures, none disappeared. The passed delta reconciles exactly: 9816 + 51 = 9867, the 51 tests added by this branch. The 31 pre-existing failures are unrelated to this work (optional Chinese NLP dependencies, brand-system assets, and a migration-guide check).

`pre-commit run --files` over the three changed files: **all hooks pass** — trailing whitespace, end-of-file, large files, case conflicts, merge conflicts, debug statements, mixed line ending, `ruff check`, `ruff format`, `bandit`, and secret detection. `mypy` reports `Success: no issues found in 4 source files` over its configured scope.

Durability is proven rather than asserted. `test_manifest_survives_every_crash_point` injects a failure at each of the five write phases (`after_write`, `after_fsync`, `before_replace`, `after_replace`, `after_directory_fsync`) and asserts **which generation is on disk at each one** — the previous manifest before the rename boundary, the new one after — plus that the crash actually fired at the expected phase and that no temporary file is left behind. A hook that never fired, or a rename that was not atomic, fails the test rather than passing vacuously.

The branch is rebased onto `2965ec10`; `git rev-list --count HEAD..2965ec10` is 0.

## Data provenance

**All test data is synthetic and generated algorithmically. No real patient data, and no real PHI of any kind, is present in this change.**

The corpus is built by a loop over four obviously-fictional sentence templates carrying deliberately fake markers, reused from the existing `tests/unit/processing/test_distributed_sharding.py` so the PHI-free assertions have something to catch if a field ever leaks. Document identifiers are formatted from the loop index (`note-000`, `note-001`, …). Shard payloads are small synthetic JSON objects; all digests, offsets and fingerprints are computed by the code under test, never hard-coded.

The PHI-free assertions validate **both keys and values** — a recursive walk yields every mapping key and every string in the structure, checked alongside `json.dumps(...)` and the bytes actually written to disk. Checking values alone would miss content that arrives as a dictionary key.

## Relationship to sibling work

This branch is the base of a five-part chain implementing #995. The following stack on it in order and will rebase as this merges:

1. #1071 — local executor and idempotent shard writes
2. #1070 — resumable execution and straggler handling
3. #1072 — Ray and Spark adapters
4. #1073 — PHI-free reports and CLI resume

The public surface above is the contract those branches build against. One behavioural constraint is worth calling out for reviewers of the follow-ups: `with_shard()` requires the replacement record to preserve `shard_id`, `fingerprint` and `document_count`, because that triple is the manifest's binding to the shard plan it was built from and is what lets a resumed run prove the corpus has not changed underneath it. Only execution state may move. The `dataclasses.replace(manifest.shard(id), ...)` pattern used throughout the tests satisfies this naturally.


maziyarpanahi:

PR #2028, opened 2026-08-02
feat: add resumable distributed batch runs with straggler recovery

### LINKED ISSUE #1070: Add resumable distributed batch runs with straggler recovery
Parent: #995

## Summary
Implement the resume entrypoint and straggler handling for distributed corpus batch de-identification runs.

## Scope
- Add resume(run_id) semantics for distributed batch runs.
- Recompute only failed, missing, or digest-invalid shards.
- Track per-shard execution counters for validation.
- Add timeout or straggler recovery behavior for stuck shard work.

## Acceptance criteria
- Resuming after 70 percent of shards complete re-executes only remaining or invalid shards.
- Killing 30 percent of local workers mid-run and resuming matches an uninterrupted run by per-document digest.
- The final manifest validates after recovery.
- No document text appears in recovery logs or reports.

Part of #995.

PouyanJay:
Closes #1070

## Relationship to the rest of OM-632 — read this first

**This PR shows 6 commits but closes one issue.** That is stacking, not scope creep.

| Commit | Issue | Owner |
| --- | --- | --- |
| `ceff36de` feat: add a durable distributed batch run manifest | #1069 | separate PR |
| `aa7e95f3` test: make the unreadable-manifest case portable to Windows | #1069 | separate PR |
| `c1414901` feat: add local executor and idempotent shard writes | #1071 | separate PR |
| `d4d3eaf3` test: stop relying on inode identity to prove atomic replacement | #1071 | separate PR |
| `2083014e` fix: keep visited nodes alive while scanning for driver-only state | #1071 | separate PR |
| **`97f946a7` feat: add resumable distributed batch runs with straggler recovery** | **#1070 — this PR** | this PR |

This branch is based on #1071, which is based on #1069, because resume planning reads the manifest #1069 defines and classifies work that #1071's executor performs. **#1069 and #1071 must merge first.** Once they do I rebase and this PR narrows to `97f946a7` alone — a single commit touching three files. Reviewing only `97f946a7` gives an accurate picture of what #1070 adds.

#1072 (Ray/Spark adapters) and #1073 (PHI-free reports and CLI resume) stack above this branch, in that order.

## Summary

Adds resume semantics and straggler classification for distributed corpus batch de-identification.

- **`resume(run_id)` semantics.** `resume_plan()` returns only the shards a restarted run must recompute, each with the reason it was selected.
- **Recomputes only failed, missing, or digest-invalid shards.** Completed shards with intact outputs are never re-run, so resuming costs what was lost rather than what was planned.
- **Per-shard execution counters for validation.** `attempts` is read to report shards that spent their budget. It is never incremented here.
- **Straggler recovery for stuck shard work**, via two independent mechanisms described below.

### This module classifies; the executor acts

Nothing here executes a shard, kills or reassigns a worker, writes to a store, or increments `attempts`. It returns `ResumePlan` and `StragglerCandidate` records and leaves every action to the executor. That separation is deliberate and is what keeps the resume logic pure and testable.

### Straggler definition

Two mechanisms, neither of which uses a wall-clock threshold:

1. **Orphan detection is clock-free.** A `RUNNING` shard whose `worker_id` is absent from `live_workers` has lost its worker. Passing `live_workers=None` treats every running shard as orphaned, which is the correct reading when the whole run died and is being resumed from its manifest. This is what covers the "kill 30% of workers" criterion, deterministically.
2. **Straggler detection uses an injected clock.** A running shard is compared against the **median per-document rate** of the shards that already finished — shards hold unequal document counts, so raw durations are not comparable. It fires above `multiplier` (default 1.5) times that median, and only once `min_completed` (default 3) shards have finished. Below that it returns empty rather than guessing from noise.

`clock` defaults to `time.time` and is injectable, so **the tests contain no `sleep` and no threads**.

### Fixed per-attempt cost

Per-document rates assume duration is proportional to document count. Real shards also pay a fixed cost per attempt, such as loading a model, which a small shard amortises over few documents — judged on rate alone, a healthy two-document shard can look slower than a healthy hundred-document one. No shard is therefore reported before it has run at least `min_elapsed_seconds`, which **defaults to the median total duration of the finished shards**. That is parameter-free, needs no arbitrary constant, and adapts to the workload. The applied floor is exposed as `StragglerCandidate.elapsed_floor_seconds` so it is auditable rather than invisible.

## Files

The issue specified scope and acceptance criteria but did not list files. All three are additive; there are **zero deletions**.

| File | Why |
| --- | --- |
| `openmed/processing/resume.py` (new) | the module |
| `openmed/processing/__init__.py` (+34/−0) | public exports, matching how `distributed.py` (#1074) and `run_manifest.py` (#1069) are surfaced |
| `tests/unit/processing/test_batch_resume.py` (new, 39 tests) | mirrors `test_distributed_sharding.py` |

### Gating tests checked, and why nothing else is touched

Each of these was checked against the actual test, not assumed:

- `scripts/release/public_api_baseline.json` — **not updated.** `tests/unit/release/test_public_api_compat.py` only fails on `breaking` changes; new exports classify as additions. #1069 set the same precedent.
- `tests/unit/release/test_import_budget.py` — requires `openmed.processing` stay lazy at `import openmed`. Satisfied: `openmed/__init__.py` is untouched.
- `tests/unit/test_no_telemetry.py` — its `ALLOWED_NETWORK_MODULES` needs an entry only for modules importing network libraries. `resume.py` imports `hashlib`, `math`, `statistics`, `time` and stdlib typing helpers only.
- `tests/unit/test_docs_publication.py` — classifies `docs/**/*.md`. **No docs added**; the CLI and operator-facing reports belong to #1073.

**No new dependencies**, so no `uv lock` change. Stdlib only.

## Findings worth carrying into review

Four things a reviewer cannot reconstruct from the diff:

1. **The odd-looking straggler fixture is load-bearing — please do not "clean it up".** A `median` → `fmean` mutant originally **survived the entire suite**, because every fixture had a per-document rate of exactly 1.0, making median and mean identical by construction. The fix was a deliberately skewed fixture with rates `[1, 1, 1, 1, 500]`: median 1.0 correctly flags a lagging shard, while mean 100.8 gives a threshold of 151.2 and misses it entirely. That skew is what makes the test non-vacuous, and it is the reason median was chosen over mean in the first place.
2. **`is_complete` returned `True` for a run that produced zero output.** With every shard `FAILED` at its attempt limit, the plan had nothing queued and nothing in flight, so a driver looping `while not plan.is_complete` would exit reporting success. A budget-exhausted run is now **terminal but not complete**; `is_exhausted` distinguishes the two. Covered by `test_exhausted_run_is_terminal_but_not_complete`.
3. **`worker_ref()` is exported deliberately and must stay shared.** Worker ids are operator-controlled free text; the manifest bounds their length and rejects control characters but does nothing to `onc-ward-dr-smith-01`. Reports therefore publish only a hashed, namespaced reference. It is a module-level exported function rather than a private helper because more than one layer publishes worker references, and two hand-rolled constructions that each look correct produce **different refs for the same worker inside a single report** — so an operator asking "which shards is this worker running?" silently gets no match. `test_worker_ref_is_a_shared_helper_not_a_duplicated_construction` pins this.
4. **`straggler_detection_enabled` is a real serialized key, not a derived property.** It duplicates information available from a null `straggler_baseline_seconds`, and that redundancy is intentional: an empty `stragglers` list is ambiguous between "nothing lagging" and "not measured", and inferring it from a Python property proved to be a footgun in practice, not merely in principle.

Two further notes for the seam with #1071:

- **`prepare_resume` deliberately preserves `output_path`, `output_digest` and `output_bytes`** on requeued shards. The executor seeds each task's `expected_digest` from the manifest's recorded digest and only compares when it is not `None`, so clearing it would silently disable the rerun-determinism check for exactly the shards a resume recomputes. An earlier revision cleared these fields; that was a real bug and is now pinned by `test_prepare_resume_keeps_the_digest_the_executor_compares_against`.
- **`reconcile_shard_attempts` raising on divergent digests depends on shard writes being genuinely content-addressed and idempotent**, which makes the executor's idempotency load-bearing for the straggler path, not only for its own.

One honest caveat: the `duration_seconds >= 0` filter in `_completed_durations` is **not individually killable by mutation**. Given the adjacent `math.isfinite` check and the manifest's own "completed_at must not precede started_at" invariant, no reachable input exercises it. It is retained as defence-in-depth and becomes live if either condition is relaxed. Flagged rather than presented as proven. Relatedly, the manifest now rejects non-finite timestamps at construction, so the tests that exercise the non-finite filter reach it through an explicit documented bypass and assert the constructor rejects the value first — they are labelled as pinning defensive behaviour, not a live path.

## Why `2083014e` is in this history

A reviewer will see a `find_driver_only_state` fix in the chain and reasonably ask why it sits under a resume PR. It was found from this branch.

After a cascade rebase, the full suite here returned **32 failures instead of 31** — the extra one being `test_shard_executor.py::test_guard_detects_every_known_smuggle_shape`, reporting that a `functools.partial` payload had evaded the driver-only guard. Since this branch's test file lives in the same chunk, this branch's commit was the prime suspect. Six experiments established otherwise, the decisive one being that the failure **reproduces with this branch's commit absent** — so it was neither caused here nor deterministic. #1071 then found the mechanism: the walker memoises nodes by `id()` while allocating temporaries (`list(partial.args)`, `dict(partial.keywords)`) that are freed as each subtree completes, and CPython reissues those addresses, so a later node inherits a memoised id and is skipped without ever being inspected. That is why it scaled with the size of the test population rather than with machine load.

This mattered because the failure is silent in the dangerous direction: the guard returns `()` — "nothing forbidden found" — for a payload it is supposed to reject. Re-running until the suite went green would have shipped that.

After the fix, the exact configuration that exposed it was re-run 3× with this branch's test file and 3× without: **6/6 pass**. The arm without the file is the one that matters, since that is where the pre-fix failure last reproduced.

## Data provenance

**All test data is synthetic and generated algorithmically. No real patient data, no PHI, and no network access at any point.** Documents are built as `{"id": f"doc-{i:04d}", "text": f"synthetic record {i}"}`; shard outputs are `sha256` of the document id. No fixtures are committed.

Reports are machine-derived throughout — shard ids, counters, timings, digests, hashed references. The single field reproduced verbatim is `run_id`, because it is the key tying a report to its manifest and the manifest already publishes it in the clear; it carries the same constraint there, that it must not embed a record-derived value. Worker ids are hashed. Failures are recorded as an exception **type name** only, never a message. `test_resume_report_is_phi_free` walks the serialized report checking **both keys and values**, using a worker id shaped like a real misconfiguration (`worker-for-doc-0007-MRN-123456`) rather than a clean one — with a clean id the test could not have caught the leak it now guards.

## Verification

Measured at base `2965ec10`, full suite:

| | passed | failed | skipped |
| --- | --- | --- | --- |
| canonical baseline `SHARED_BASELINE_2965ec10_FAILED.txt` | 9816 | 31 | 86 |
| this branch (`97f946a7`) | **9955** | **31** | 86 |

**The sorted FAILED sets are identical**, and the count is a stable 31 across repeated runs — not the transient 32 described above. All 31 are pre-existing and environmental (optional OpenCC/jieba dependencies, brand system, sigstore, MCP entrypoint, API-surface diff). No test regresses.

The +139 reconciles exactly against the six stacked commits: 51 (`test_batch_run_manifest.py`, #1069) + 49 (`test_shard_executor.py`, #1071) + **39** (`test_batch_resume.py`, this PR) = 139, and 9816 + 139 = 9955.

`pre-commit run --files` on all three changed files: **all hooks pass** (ruff check, ruff format, bandit, secret detection, EOF, whitespace, merge-conflict, debug statements).

### Windows portability audit

The two files here were audited against the four traps that broke Windows CI elsewhere in this chain. All four are confirmed absent, stated as positive evidence rather than a bare "none found":

- **Permission bits** — no `chmod`, `S_IREAD`, `S_IWRITE` or `os.access` anywhere. No test makes a file unreadable; the digest-mismatch path is exercised by *overwriting* content, which is portable.
- **Path-string comparison** — no `os.sep`, no `str(Path…)`, no `os.path.join`, and no comparison of a path against a literal. Recorded output paths are `Path.name` only, and `resume.py` compares digests and shard ids, never path strings.
- **Encoding** — `resume.py` performs **no file I/O at all**. All six text calls in the test file pass `encoding="utf-8"` explicitly.
- **`st_ino`** — no `st_ino`, `samefile`, `st_dev` or file-identity check of any kind, and no `os.name` / `sys.platform` branching. Rewrites are proved by **content and digest**, which is portable and also a stronger claim.

**One deliberate judgment, so it is not mistaken for an oversight:** `write_text` uses platform-default newlines, so shard outputs contain `\r\n` on Windows. That is safe here because every digest and byte-count comparison is *within-run and within-platform* — `output_bytes` comes from `st_size` at write time and is only ever compared to itself, and the `'0' * 64` digest fixtures are never compared against a digest computed from real content. No assertion depends on cross-platform byte identity.

**Windows CI context:** `test_budget.py::test_wall_time_clock_raises_after_deadline`, seen failing on sibling PRs, is pre-existing upstream code present at `2965ec10` and absent from every diff in this chain. Not ours.

## Acceptance criteria

| Criterion | Proving test |
| --- | --- |
| Resuming after 70% of shards complete re-executes only remaining or invalid shards | `test_resume_recomputes_only_incomplete_shards` — 10 shards, 7 complete, resume set is exactly `(7, 8, 9)` |
| Killing 30% of local workers mid-run and resuming matches an uninterrupted run by per-document digest | `test_interrupted_run_matches_uninterrupted_by_document_digest` — 3 of 10 shards orphaned, recovered, per-document digest map compared against an uninterrupted run |
| The final manifest validates after recovery | `test_manifest_validates_after_recovery` — every shard completed, all recorded outputs revalidated |
| No document text appears in recovery logs or reports | `test_resume_report_is_phi_free` — serialized report walked over keys and values |

Every test above was verified to be falsifiable by mutation. For the second criterion in particular: making resume skip orphaned running shards leaves the resume set empty, recovers **23 of 40 documents**, and the digest maps diverge — so the test measures recovery rather than merely completing without error.


maziyarpanahi:

Issue #995, opened 2026-06-28
Build a distributed corpus-scale batch de-identification engine with sharding and resumability


maziyarpanahi:
## Summary
BatchProcessor in processing/batch.py runs in a single process and re-does all work on failure, so de-identifying a multi-million-document corpus is neither horizontally scalable nor resumable. Operators need to shard a corpus across worker processes or a cluster, tolerate worker loss without losing or duplicating documents, and resume an interrupted run from a durable manifest while keeping every model inference on-device.

## Scope
- [ ] Add a deterministic content-addressed sharding layer that partitions a document iterable into stable shards by a stable document id hash, independent of input ordering and worker count.
- [ ] Introduce a durable run manifest (shard -> status -> output digest) with atomic per-shard commit so completed shards are skipped on resume and partial outputs are never double-counted.
- [ ] Define an Executor protocol with a local multiprocessing backend and optional Ray and Spark adapters (lazily imported) that map shards to workers and reduce per-shard summaries.
- [ ] Add at-least-once-with-idempotent-write semantics: each shard writes to a temp path and atomically renames on success, so a re-executed shard overwrites identically rather than appending.
- [ ] Implement straggler handling and a resume(run_id) entrypoint that recomputes only failed or missing shards and validates output digests against the manifest.
- [ ] Emit a PHI-free aggregate run report (per-shard counts, durations, failures, fingerprint) and a CLI/programmatic resume path.

## Acceptance criteria
- [ ] Killing 30 percent of workers mid-run and resuming produces an output set identical (by per-document digest) to an uninterrupted run, with each document de-identified exactly once.
- [ ] Throughput scales near-linearly to at least 6x on 8 local worker processes versus single-process BatchProcessor on a 100k synthetic-document benchmark (>=5x effective speedup).
- [ ] resume(run_id) after completing 70 percent of shards re-executes only the remaining shards (verified by per-shard execution counters) and the final manifest validates.
- [ ] Ray and Spark adapters pass the same sharding/resumability conformance suite as the local backend behind the Executor protocol.
- [ ] All model inference runs on-device and no document text appears in the manifest, digests, or run report (scanner test).

## Out of scope
- Cluster provisioning, autoscaling, or scheduler deployment.
- Distributed shuffle/joins beyond shard map-reduce.
- Streaming ingestion (covered by the streaming connector task).

## Files
- openmed/processing/batch.py
- openmed/processing/__init__.py
- openmed/processing/distributed.py
[...]

PR #2029, opened 2026-08-02
Add local executor and idempotent shard writes for distributed batch runs

### LINKED ISSUE #1071: Add local executor and idempotent shard writes for distributed batch runs

PouyanJay:
Closes #1071

## Relationship to sibling PRs — read this first

**This branch contains six commits but closes one issue.** That is expected, not scope creep:

| Commit | Issue | Owner |
|---|---|---|
| `ceff36de` | #1069 — durable distributed batch run manifest | sibling PR, **must merge first** |
| `aa7e95f3` | #1069 — Windows portability of its unreadable-manifest test | sibling PR, **must merge first** |
| `c1414901` | #1071 — the executor | **this PR** |
| `d4d3eaf3` | #1071 — stop proving atomic replacement via inode identity | **this PR** |
| `2083014e` | #1071 — fix `id()` recycling in the driver-only-state guard | **this PR** |
| `c563419f` | #1071 — retry shard publication when a duplicate writer wins the race | **this PR** |

#1071 builds directly on #1069's `openmed/processing/run_manifest.py`: this executor consumes `BatchRunManifest`, `ShardRecord`, `ShardStatus`, `RunManifestStore`, `shards_to_execute` and `ShardOutputDigestMismatchError`, and it defines none of them. Reviewing #1069 first will make this diff far easier to read.

**Rebase promise:** when #1069 merges, I rebase this branch onto the new `master` and this PR narrows to my four commits over three files. No action needed from a reviewer; ping me and I will rebase on request.

This is the third of five stacked branches for parent #995. #1070 (resumable runs / straggler recovery), #1072 (Ray and Spark adapters) and #1073 (PHI-free reports and CLI resume) stack **above** this one and depend on the executor contract landed here.

## Summary

Adds the executor abstraction and a local backend that runs shard work with at-least-once execution and idempotent writes.

- **`ShardExecutor` protocol** — the contract distributed batch processing dispatches through. #1072's Ray/Spark adapters implement this same protocol.
- **`LocalShardExecutor`** — sequential at `max_workers=1`, a thread pool above that, and a process pool whenever `use_processes` is set.
- **Atomic shard publication** — each shard is written to a same-directory temporary file, fsynced, renamed over the destination, and the directory fsynced, reusing `batch.py`'s existing `_atomic_write_bytes`.
- **Run-level reduction** — per-shard summaries reduce into `ShardRunResult`.

### The division of responsibility that makes it safe

Workers run the handler, publish the output file, and return **PHI-free metadata only** — shard id, status, digest, byte count, duration, worker id, and an exception *type* name. The driver is the **sole writer of the run manifest**: it increments `attempts` at dispatch and records the outcome when a worker reports back. Payloads never travel back across a process boundary, so no document text can be pickled into the driver, and no two writers ever race on manifest state.

## Deviations from the issue

The issue listed no files, so every file below is stated explicitly with the reason it exists.

| File | Why |
|---|---|
| `openmed/processing/shard_executor.py` | New. The executor, the atomic write path, and the run-level reduction — the issue's four scope bullets. |
| `openmed/processing/__init__.py` | Forced by `tests/unit/test_public_api_docstrings.py`, which resolves every name in the public export inventory and requires a meaningful docstring on each exported function/class. **+30 insertions / 0 deletions** — a pure union, so the three sibling branches that also extend this file do not conflict. |
| `tests/unit/processing/test_shard_executor.py` | New. 51 tests. |

Three files, mirroring the structure of the merged sharding PR #1074 (`9c62a9ed`), which was also module + exports + one test file.

**Deliberately not added:** no `docs/*.md` (would enrol in the docs-publication classification gate), no CLI (that is #1073's), no changelog entry, and no new dependency — `concurrent.futures` is stdlib, so `uv.lock` is untouched.

**Deliberately left to siblings:** retry policy, straggler detection and speculative execution (#1070); Ray and Spark backends (#1072); PHI-free reports and the CLI resume path (#1073). A shard that fails here is simply eligible again on the next run; this module classifies nothing and reassigns nothing.

## Idempotence contract

Re-running a completed shard produces byte-identical output and never corrupts a partially-written one.

- **Mid-write crash** — the final path holds either the previous bytes or the complete new bytes, never a partial write. The temporary file is unlinked on any exception. A shard is marked `COMPLETED` only *after* the rename returns, so a crash in between leaves it `RUNNING` and the next run re-executes it.
- **Duplicate concurrent writer** — expected under at-least-once delivery. Each writer gets its own temporary file and the rename is atomic, so the loser's bytes are discarded whole rather than interleaved.
- **Output exists but the digest mismatches** — the shard is re-executed and the file atomically replaced. Re-execution overwrites the same final path; it never appends.
- **Partial temporaries** — dotfile-prefixed with a `.tmp` suffix, never named in the manifest, and never counted as a completed output.

Atomicity assumes `os.replace` on a local POSIX filesystem or NTFS. It is **not** guaranteed atomic on NFS or SMB, so shard output roots should be local storage — the same assumption `batch.py`'s checkpointed writer already makes. Output roots must also be run-scoped: two runs sharing a `root` overwrite each other's shard files and neither converges. Both constraints are documented in the module.

## Findings from adversarial review that the diff does not show

These were real defects found and fixed during review, not hypotheticals. Each is recorded because the reasoning is not recoverable from reading the final code.

### 1. The driver-only-state guard was bypassable by five shapes, and one of them crashed the driver

A worker must never receive the manifest or its store. Both pickle **without error** and round-trip to a *different* object, so a worker mutating its copy would report success while the driver's manifest stayed untouched. `BatchRunManifest` is the nastier case: it round-trips comparing **equal but not identical**, so even a defensive equality assertion passes at the moment of transport and the copy only diverges once mutated. That is why the guard is structural rather than a pickle check.

The first implementation walked only `__dict__`, closures, partials and bound methods. It was evaded by **`__slots__`, class attributes, default arguments, `@property`**, and it was too shallow — a store nested more than four levels behind a partial keyword was missed. Worse, probing a payload with a permissive `__getattr__` (a config proxy, a lazy wrapper, a `MagicMock`) **crashed the submission path on payloads containing no store at all**.

Both are fixed: the walker now covers all five shapes, depth is 12 with a node budget, every attribute probe is exception-guarded, and closure cells are type-checked before access. Properties are inspected **structurally via `fget` and never invoked** — the guard must not run user code to decide whether to run user code.

The guard is **defence in depth, not the primary control**. The structural fix is the one this module uses internally and that callers should copy: submit a module-level worker function over plain data, so nothing is reachable to smuggle. A walker can always be out-recalled by a sufficiently indirect capture; a worker holding no reference cannot be.

The custom-store check is a documented **heuristic**: it matches a type defining both `load(self)` and `save(self, manifest)`. A `load` taking no arguments beyond `self` is what makes it specific. An unrelated captured object exposing that same pair would also be rejected — the intended direction to be wrong in, since a named error at submission time is recoverable and silently discarded manifest updates are not. `isinstance` against the protocol was rejected deliberately: `runtime_checkable` verifies attribute *presence*, not signatures, and would be strictly broader.

### 2. The guard itself returned "clean" for payloads it had to reject — `id()` recycling

**This was a live defect in the safety path, not a test problem, and it is the most important thing in this PR.**

`find_driver_only_state` memoized visited nodes in a `set[int]` keyed on `id()`, without holding a reference to the objects. An id is only unique among **live** objects — and the walker allocates its own temporaries as it goes, `list(partial.args)` and `dict(partial.keywords)`, each freed the moment its subtree completes. CPython reissues that address to the next allocation, so the next node inherits a memoized id and is **skipped without ever being inspected**. The walker was poisoning its own memo.

The consequence is the dangerous direction: `find_driver_only_state` returns `()` for a payload carrying the manifest or its store, and the executor admits it to a worker.

It scales with the number of same-shaped nodes, which is what separates a structural bug from a flake. Store placed in the **last** partial, so it is found only if no earlier temporary poisoned the memo:

| partials ahead of the store | misses / trials (before the fix) |
|---|---|
| 0, 1, 2, 5, 20 | 0 / 200 |
| **100** | **200 / 200** |
| **500** | **200 / 200** |

After the fix: **0 / 200 at every size.** As a regression test it fails **10 of 10** runs without the fix and passes 10 of 10 with it, so the reproduction is deterministic rather than a stress run.

The fix holds every visited node alive for the duration of the walk, so no id can be recycled mid-scan. The memo's original purpose was re-verified in a companion test, because fixing reuse must not amount to switching the memo off: cycles still terminate, a node reached by two paths is still reported exactly once, and shared subgraphs stay bounded.

**Credit:** this surfaced because #1070 saw 32 failures instead of 31, treated **its own commit as the prime suspect**, and reproduced the failure with its commit absent — rather than re-running until the suite went green. The test passed in isolation, passed across all 47 tests in this file, and passed 4/4 in a chunk that included #1070's file; it failed 1/3 only with that file removed. Re-running would have buried it.

I own the miss as well as the fix: my `st_ino` change one commit earlier identified this exact both-directions failure mode — an identity check that fails a correct implementation in one direction and silently passes a broken one in the other — and I did not think to audit my own `id()` memo for the same shape.

### 3. The digest guard fired after publishing and permanently wedged the run

The mismatch comparison originally ran in the driver **after** the worker had already renamed the new bytes over the old. A non-deterministic handler therefore destroyed the good output *and* raised, and because the raise happened mid-loop it also abandoned shards that had already succeeded. Under `force=True` on a healthy run this corrupted every shard.

Now the worker compares the recomputed digest **before** the rename, so existing output is never replaced; every shard is still recorded and persisted; and the error is raised once at the end, after the manifest is consistent. A failed attempt also **preserves** the `output_digest` it published earlier — clearing it would let one transient `IOError` launder a later non-deterministic result past the comparison.

### 4. Concurrent duplicate publication failed the shard on Windows

Found by CI on this PR, on a supported platform, in the documented normal path.

`os.replace` on Windows refuses `MoveFileEx` with a sharing violation while another handle holds the destination, so when two workers publish the same shard at the same moment the loser's replace raises `PermissionError` and the shard fails. POSIX `rename(2)` has no such restriction, which is why every non-Windows run was green.

This is not a corner case for this design. At-least-once execution means duplicate publication is **expected**: #1070's attempt reconciliation and #1072's speculative execution both rely on a duplicate writer being safe.

**Availability, not corruption** — and that distinction is from the data, not inference. Injecting the failure reproduced the CI symptom exactly:

| trial | writers that completed | errors | file intact |
|---|---|---|---|
| 0 | 3 / 8 | `PermissionError` | **yes** |
| 1 | 2 / 8 | `PermissionError` | **yes** |
| 2 | 1 / 8 | `PermissionError` | **yes** |

One complete payload survived every time. So the guarantee that held is *"one complete payload survives"*; the guarantee that broke is *"the loser's write does not fail the shard."*

The fix is a bounded retry with backoff that republishes **in full from a fresh temporary** each attempt, so a retry can never append or strand a partial file. The bound is deliberate: once spent, a genuinely unwritable destination raises `PermissionError` unchanged rather than being mistaken for a lost race. Post-fix the same injection gives 8/8 with zero errors, under an emulation stricter than real Windows.

Both halves of that contract are pinned, which two mutations killing *different* tests demonstrates:

| Mutation | Test it kills |
|---|---|
| Remove the retry (pre-fix behaviour) | `test_publish_survives_a_windows_style_sharing_violation` |
| Swallow `PermissionError` instead of raising after the bound | `test_persistent_permission_error_is_not_mistaken_for_contention` |

Both tests run on **every** platform by injecting the error rather than being Windows-gated, so the contract is now verified on Linux and macOS too — not only where it happened to break.

**A separate testing defect made this much harder to diagnose, and is worth stating on its own.** An exception raised inside a `threading.Thread` never reaches the test body, so a writer that died was indistinguishable from one that never ran. A real product bug therefore surfaced as a bare `assert 4 == 8` rather than the `PermissionError` that caused it. The threads now capture exceptions and assert on them first. This lesson is independent of the bug that exposed it: any test that fans work out to threads and asserts on a collected result needs to assert on collected *failures* first.

### 5. This module's idempotency is load-bearing for two other branches

Flagged explicitly because it is the kind of coupling a good-faith refactor breaks:

- **#1070's `reconcile_shard_attempts`** depends on `attempts` incrementing exactly once per dispatch — and not at all when a resumed run finds every output still valid.
- **#1073's resume path** depends on `prepare_resume` preserving `output_digest` across a failed attempt.

Both are verified on this exact tree: after a first run `attempts=[1]`; after a resume with all outputs valid `attempts=[1]` with zero executions; after `force` `attempts=[2]`. After an all-failing forced attempt, every `output_digest`, `output_path` and `output_bytes` is preserved and the bytes on disk are untouched.

### 6. Shard metadata must not be routed through the input-normalization gateway

`openmed/processing/` does not import `openmed/utils/gateway.py` and this PR does not add such an import. Recording the reason so it stays that way: `normalize_text()` funnels into `_coerce_to_text` (`gateway.py:139-146`), which deliberately calls `str(value)` on non-string input. That coercion is correct for document *text* payloads and wrong for identifier-class fields — a dict arriving in a metadata field would come back as its rendered string and be written to disk.

Accordingly, `_safe_worker_id` and `_safe_error_type` **reject and substitute a constant** rather than coerce, matching the gateway's own `validate_language`, which requires `isinstance(str)` and refuses to coerce.

`_safe_error_type` also exists for a concrete reason: an exception class built dynamically, or a Ray-style `RayTaskError(ValueError)` wrapper, is not a valid dotted identifier and the manifest rejects it. Passing it straight through would crash the driver *while it was recording the failure*, turning a recoverable shard error into a lost run.

## PHI safety

Manifest and result records carry no document text, no document identifiers, no document hashes, and no exception messages — only an exception **type**. Shard file names are derived from the shard id alone (`shard-{id:05d}.jsonl`), never from a document id, which matters because `output_path` is written into manifest bytes.

This is enforced, not asserted: the PHI tests scan the **serialized** form and check **both keys and values recursively**, because content riding through as a dict key is invisible to a value-only scan. A mutation that makes shard names document-derived fails 12 tests.

## Data provenance

**All test data is synthetic and generated algorithmically.** Document identifiers are produced by format string (`note-{i:04d}`), text is templated (`Synthetic subject {i:04d} reached extension 555-{i:04d}`), and payloads derive from `stable_document_hash`. No real patient data, no real PHI, and nothing derived from a real corpus. Phone-style numbers use the reserved `555-` prefix.

## Verification

|---|---|---|---|---|
| Canonical baseline at `2965ec10` | 31 | 9816 | 86 | 1 |
| This branch (`c563419f`) | 31 | 9918 | 86 | 1 |

**The sorted FAILED sets are identical — zero new failures and zero pre-existing failures fixed.** The 31 are pre-existing on unmodified upstream (optional Chinese NLP dependencies, brand-asset portability, and similar) and are unrelated to this change. The +102 passed is exactly `51` (#1069's tests, including its Windows portability commit) `+ 51` (this PR's).

`pre-commit run --files <the three files>`: **all hooks pass**, including `ruff check`, `ruff format`, `bandit` and the secret scan.

### Mutation testing

Every safety property was verified by deliberately breaking the implementation and confirming a test fails. All 21 mutations were killed:

| Mutation | Kills |
|---|---|
| Crash hook ignored in the worker | 7 |
| Non-atomic in-place write | 2 |
| `force` never re-dispatches | 8 |
| `error_type = str(exc)` | 9 |
| Empty-shard handler bypass removed | 1 |
| Chunked non-atomic write (concurrency) | 1 |
| Document-id-derived shard filename | 12 |
| Driver-only guard removed | 2 |
| Structural store check dropped | 2 |
| Walker reduced to `__dict__` only | 1 |
| Unguarded closure access | 4 |
| Digest compared after writing | 2 |
| Digest cleared on a failed attempt | 1 |
| Guard moved after dispatch bookkeeping | 1 |
| `max_workers=1` short-circuits past `use_processes` | 1 |
| Raw `type(exc).__name__` | 1 |
| Thread path left ungated | 1 |
| Visited nodes not held alive (`id()` recycling) | 1, deterministically 10/10 |
| Inode gate forced off, then non-atomic write / no-op re-dispatch | 1 and 3, proving the portable assertions carry |
| Retry removed (pre-fix Windows behaviour) | 1 |
| `PermissionError` swallowed instead of raised after the bound | 1 |

Two of these caught tests that were passing **for the wrong reason** and were rewritten:

- The empty-shard test used a handler returning `b""` regardless, so it could not distinguish the bypass from correct behaviour. It now uses a handler that raises if handed an empty shard.
- The concurrent-writer test had all eight writers write *identical* bytes, making tearing undetectable by construction. It now uses eight distinct equal-length payloads and asserts the survivor equals exactly one of them.

Stated honestly: that concurrency test is **probabilistic (~3 in 5) against a broken implementation**, because a non-atomic write tears only when threads actually interleave.

**Correcting a claim made earlier in this PR's own history:** I originally described that test as *deterministic against a correct implementation*. That was true only on POSIX. On Windows the correct implementation raised `PermissionError` from `os.replace` under exactly the contention the test creates, so the test failed against code that was behaving as designed — see finding 4. The gap was invisible to me locally and Windows CI is what exposed it. This is the same failure mode this PR keeps documenting elsewhere: a check that appears to pin a property while actually depending on an unstated platform assumption.

### Acceptance criteria

| Criterion | Test |
|---|---|
| Local execution matches single-process digests | `test_threaded_execution_matches_sequential_digests`, `test_process_pool_execution_matches_sequential_digests` |
| Re-executing overwrites rather than appends | `test_reexecuting_completed_shard_overwrites_same_bytes` — proves the overwrite path actually ran by instrumenting the atomic-write hook, not merely by comparing bytes |
| Partial temporaries are not counted | `test_crash_during_shard_write_never_leaves_a_partial_output` (all 5 atomic-write phases), `test_planted_temporary_file_is_never_counted_as_a_shard_output` |
| Worker failures are covered | `test_worker_failure_records_error_type_without_leaking_detail`, `test_failed_shard_is_reexecuted_and_completes`, `test_unnameable_exception_type_does_not_crash_the_driver` |

The crash tests are parametrized over all five atomic-write phases, and a companion test asserts the on-disk state genuinely **differs** across the rename boundary — previous bytes before `os.replace`, new bytes after. Without it, a hook that silently no-opped would pass every phase.

## Windows portability

Two things here, one mine and one not.

**Mine (`d4d3eaf3`).** Three assertions proved atomic replacement by comparing `st_ino`. That is not portable: `st_ino` is 0 on some filesystems, including Windows outside NTFS, and it breaks in **both directions at once** — "the inode changed" becomes trivially false and *fails a correct implementation*, while "the inode is unchanged" becomes trivially true and *passes a broken one silently*.

The product contract is now proved portably everywhere — the overwrite path by instrumenting the atomic-write hook, non-destructive re-execution by comparing bytes — and only the corroborating inode assertions are gated. The gate is a **capability probe** (write two files, check for distinct non-zero inodes), not an `os.name` guess, so it also covers non-NTFS mounts on POSIX rather than assuming the problem is Windows-shaped.

Verified by forcing the probe false: a correct implementation still passes, and both a non-atomic write and a no-op re-dispatch are still caught. Honest caveat — on such filesystems the non-atomic-write mutation is caught by **one** test instead of two. Redundancy is lost, the guarantee is not.

**Also mine (`c563419f`).** Windows CI found a genuine product defect in concurrent shard publication; it has its own section above as finding 4, because it is a defect on a supported platform rather than a test-portability issue.

**Not mine.** `tests/unit/core/test_budget.py::test_wall_time_clock_raises_after_deadline` fails on Windows across sibling PRs in this stack. It is **pre-existing upstream code**, present at `2965ec10` and absent from every one of our diffs: Windows `time.monotonic()` has ~15.6 ms granularity, against a `max_wall_time=0.001` deadline. Noted so a reviewer scanning red checks across the batch can tell which failures this stack owns.

## Empty shards

A shard with no documents bypasses the caller's handler and publishes an empty output, settling as `COMPLETED`. This is deliberate. #1069 requires `output_path` and `output_digest` for a `COMPLETED` record, so simply filtering empty shards would leave them `PENDING` forever, causing `shards_to_execute` to re-queue them on every run — where #1070 would read them as permanent stragglers. An empty file is also what distinguishes "ran, produced nothing" from "never ran".

This interacts with fixtures: with 12 documents over 5 shards there are no empty shards, but with 6 over 3 there is one. Tests that assume every shard runs the handler now assert that shape explicitly rather than trusting the hash distribution to hold.


maziyarpanahi:

PR #2030, opened 2026-08-02
feat: add PHI-free run reports and a batch-run CLI

### LINKED ISSUE #1073: Add PHI-free reports and CLI resume path for distributed batch runs

PouyanJay:
Closes #1073

## Relationship to the rest of the batch chain — read first

This branch sits at the top of the `#995` distributed-batch chain and **contains its four predecessors' commits**. Ten commits, two of them mine:

| | branch | commits |
| --- | --- | --- |
| #1069 | durable run manifest | `ceff36de`, `aa7e95f3` |
| #1071 | local executor, idempotent shard writes | `c1414901`, `d4d3eaf3`, `2083014e` |
| #1070 | resumable runs, straggler recovery | `97f946a7` |
| #1072 | Ray and Spark adapters | `f7782006`, `cd5e5443` |
| **#1073 (this PR)** | PHI-free reports, `batch-run` CLI | `91c0b193`, `6e581549` |

**Merge order: #1069 → #1071 → #1070 → #1072 → this PR.** Everything below `#1073` belongs to those PRs and should be reviewed there; only the last two commits are new here.

**Rebase promise:** if any predecessor is amended or lands in a different order, I will rebase this branch onto the resulting tip with `git rebase --onto <new-tip> <old-base>`, re-verify that only my two commits replay, and re-run the full suite before re-requesting review. Base is `2965ec10`; `git rev-list --count HEAD..upstream/master` is `0`.

## Summary

Reduces a run manifest — and optionally a resume plan and an output validation — to an operator-facing report carrying per-shard counts, attempts, durations, failures and fingerprints, and adds `openmed batch-run start|resume|report` over the existing local executor.

- Aggregate per-shard counts, durations, failures and the run fingerprint.
- A CLI path to start or resume a distributed run by run id.
- Scanner coverage proving no document text reaches manifests, reports or logs.
- The local-first operational workflow, documented.

## Design: the report re-derives rather than forwards

A manifest and a resume plan are internal structures whose fields are chosen for the driver that owns them. A report is an artifact an operator files and pastes into a ticket, so it is held to a stricter standard than either source. Two fields show why:

- **`worker_id` is never published.** The manifest bounds its length and rejects control characters, which does nothing to a host named after the ward it serves. Reports carry `worker_ref` (`resume.worker_ref`, the same helper both paths call, so one worker yields one token wherever it appears).
- **`output_path` is never published.** A relative path with no traversal is safe to *store*; a filename is chosen by whoever configured the run and is not worth putting in a pasted artifact.

Payloads are validated against an allowlist before being returned, and the check covers **mapping keys as well as values** — a sentence lifted out of a record reaches a payload as easily as a key as it does as a value, and a values-only check reads as thorough while missing half the surface.

## Three things worth stating plainly

**1. The guard governs publication, not persistence.** This module sees a manifest only once it exists, so it can refuse to publish a bad value but cannot prevent one being written. That distinction is exactly what made the `--run-id` bug possible: validation fired inside `to_dict()`, *after* `store.save()`. `is_publishable_token` is exported so callers apply the **same predicate** at the point of acceptance — two regexes that drift is how the bug returns, so a test asserts the two agree in both directions across eight values including the empty string, a trailing newline and a 129-character id.

**2. Why the per-record document hash was removed.** An earlier revision wrote `stable_document_hash(document_id)` on every shard output line. A fixed-namespace digest over a low-entropy identifier — medical record numbers, accession numbers — is recoverable by enumeration, so it was a reversible pseudonym rather than a protection. It was also *new* persistence: `build_run_manifest` stores only a set fingerprint over document hashes and never per-document values, so nothing else in the chain kept a per-record token. Output lines now carry `{"text": ...}` only. Records stay joinable because sharding is deterministic — line `n` of a shard is document `n` of that shard.

**3. `error_type` depends on #1069's `isinstance(str)` fix.** This layer normalises a non-identifier exception name to a constant, but an identifier-shaped value passes unchanged, and #1069's constructor *raises* rather than repairs — validation is the failure mode, not the protection. I have no independent defense for that field and rely on the manifest constructor.

### The walker inspects every node, and that was probed rather than asserted

`assert_no_raw_text` recurses over arbitrary nested payloads checking keys and values, and **it memoizes nothing**. That is deliberate. Recording visited nodes by `id()` is the obvious optimisation and it is unsound for a walker that frees each temporary as it advances: the interpreter reissues the address, and the next node inherits a memoized id and is skipped without ever being inspected.

Two sibling branches in this chain shipped exactly that defect — #1071's guard returned `()` for payloads it must reject, and #1072's unwrapper lost the cause from depth 5. In a guard whose job is to find record content, the same defect means **the content is published**.

A grep showing no `seen` set is not evidence the walk is sound, so it was probed with the shape that surfaces the bug: a `Mapping` that constructs its members **per access**, so each temporary is collectable the moment the walk moves on, with the planted value in a **late** node rather than a shallow one.

| shape | result |
| --- | --- |
| depth 100, planted at 1 | caught |
| depth 200, planted at 3 | caught |
| depth 400, planted at 7 (~393 levels deep) | caught |
| depth 60, planted at 0 (deepest node) | caught |
| 500 short-lived siblings, planted last (500/500 distinct ids observed) | caught |

Committed as `test_guard_inspects_every_node_when_temporaries_are_freed_as_it_walks`. The reasoning is recorded in the function's docstring so the optimisation is not added later:

> Re-inspecting a shared subtree costs a repeat visit; skipping one costs the guarantee.

### A known limit, with its qualification

`algorithm` and `openmed_version` are free-form single tokens with no checkable shape, so the allowlist admits them: `ward-3-dr-quantum` passes on its face. **Reaching this requires a caller fabricating their own manifest** — the test covering it needs `object.__setattr__` to construct the state at all, because the normal constructor and every code path in the chain populate these from package constants. It is self-inflicted, not a route by which a normal run leaks. If a maintainer wants it closed, the fix belongs in `run_manifest.py`'s constructor, not here: adding a format constraint from this branch would give #1069's constructor and this guard a shared assumption neither file states.

Digest-shaped fields *are* checked (`fingerprint`, `plan_fingerprint`, `output_digest`), which is what lets a value coerced to `"None"` upstream be caught here.

## CLI

`openmed batch-run start|resume|report`, all accepting `--json` and emitting exactly **one** envelope.

Exit codes follow the repository's gate convention: `0` when every shard finished with a usable output, `1` when shards remain outstanding or have spent their attempt budget, `2` for usage. **A run that produced nothing never exits `0`.** `--max-attempts` makes an exhausted run reachable so a scripted resume loop terminates against a deterministically failing shard.

`--run-id` is validated before anything is written, because a run id that reaches the manifest cannot be withdrawn and every later `report` and `resume` re-reads it — a single space would otherwise process the whole corpus, fail at render time, and leave the run unreadable with no recovery short of editing `manifest.json` by hand.

### `--pattern` case-folding makes the corpus platform-dependent

`pathlib` defaults `case_sensitive` to false on Windows, so the same directory yields a different document set there:

```
posix default    matched=['note.txt']              stems=['note']
WINDOWS default  matched=['NOTE.TXT','note.txt']   stems=['NOTE','note']
```

Because file stems are the document ids, a different set means a different shard plan and therefore a **different `plan_fingerprint`** — which matters precisely because this PR presents that fingerprint as a reproducibility anchor. **Fingerprints are comparable across runs on one platform, not across platforms.**

Nothing is lost or collides: the stems are distinct strings, each maps to its own shard, and duplicate detection does not falsely fire. So this is documented rather than fixed by pinning `case_sensitive=True` — forcing it would make a Windows user's `.TXT` corpus silently vanish behind "No input files matched", and `--pattern` is the caller's to choose. The workaround is in the docs: give the pattern an explicit case, or normalise extensions beforehand, when a run has to reproduce on both.

Verified by probe, not by reading the docs — and worth noting the filesystem is not the variable: this was measured on a case-insensitive filesystem where `*.txt` still did **not** match `NOTE.TXT`, because glob pattern-matching is case-sensitive independently of the filesystem.

## Files changed, and why each is here

Issue-listed scope is the report module, the CLI and the docs. Everything else is forced by a gate:

| File | Why |
| --- | --- |
| `openmed/processing/run_report.py` | New: the report module. |
| `tests/unit/processing/test_batch_run_report.py` | New: 24 tests. Named to mirror #1069's `test_batch_run_manifest.py`. |
| `tests/unit/cli/test_batch_run_cli.py` | New: 20 tests. |
| `openmed/processing/__init__.py` | Export block. Appended as its own alphabetically-sorted group, touching no other line, so the chain's four-way edit resolves as a pure union. |
| `openmed/cli/main.py` | Command group and handlers. |
| `docs/batch-processing.md` | **Appended to an existing page rather than adding a new one** — `tests/unit/test_docs_publication.py:171` asserts `set(navigated) \| set(link_only) == _public_default_markdown()`, so a new `docs/*.md` would require `mkdocs.yml` nav and `publication.yml` edits. This page is already in the nav (`mkdocs.yml:59`). |

Auto-enrolling gates that constrained the work without adding files: `tests/unit/cli/test_cli_json_output.py:55` (every leaf needs `--json` + `command_path`, satisfied by the existing `_finalize_parser` walk) and `:205` (AST scan forbidding interactive input anywhere under `openmed/cli`, so resume is non-interactive); `tests/unit/cli/test_cli_help_coverage.py` (help text on every parser, and one parametrized case per command path).

**Deliberately not touched, verified rather than assumed:** `openmed/interop/tools.json` and `MCP_TOOL_NAMES` are driven by `openmed.mcp.tool_registry.TOOL_REGISTRY`, not by the CLI, so `tests/unit/interop/test_tool_schema_sync.py:124` does not fire — no MCP tool is added. `docs/api/openapi.json` is derived from the FastAPI app; no REST route is added. `tests/unit/test_public_api_docstrings.py` gates only top-level `openmed.__all__`, which the sibling sharding module also stays out of.

### Documentation placement

`docs/batch-processing.md` merged cleanly with #1072 only because we used different anchors — my section landed mid-page, #1072's at EOF, leaving ~350 unrelated lines between two distributed sections. I moved #1072's section up to sit immediately after mine, giving one progression: *crash-safe checkpoints → distributed runs (manifest, resume, reporting) → distributed shard execution (backends) → operations*. Verified as a **pure permutation** by diffing sorted content against the pre-move file, which caught and removed one stray blank line my splice introduced. #1072's prose is unmodified; the only new text is a one-line forward pointer in my own section.

## Data provenance

All test data is synthetic and algorithmically generated. Notes are built from an index (`f"Note {index:05d}..."`), document ids are file stems derived from the same index, and shard failures are injected by a deterministic predicate rather than randomness. PHI strings are invented (`Evelyn Quantum`, `ZQ-7391`, `042-66-9001`) and reuse the vocabulary already present in the repository's `tests/unit/test_no_raw_text_logging.py` guard. **No real patient data, no corpus, no network access.**

## Verification

Canonical baseline **31 failed / 9816 passed** at `2965ec10`. This branch: **31 failed / 10098 passed / 88 skipped / 1 xfailed**. FAILED sets are **identical**, diffed clean against the shared baseline.

Reconciliation against an earlier chain tip, measured in this worktree rather than inherited:

```
2965ec10 (canonical)          9816 passed,  86 skipped
  + chain-added tests         +224 passed,  +2 skipped   = +226 collected
197f76e2 (chain tip)         10040 passed,  88 skipped   [measured independently]
  + 24 report tests + 20 CLI tests                        +44
  + help-coverage 82 -> 86 (one case per new command path) +4
                             10088 passed,  88 skipped
```

The `+226` is derived, not assumed: those five chain commits touched nine test files collecting **229**, of which `test_core_does_not_import_adapters.py` contributes **0** — #1072 modified it without adding test functions (3 tests before and after), so those 3 were already inside the 9816. `229 - 3 = 226`. The 2 skipped are #1072's Ray/Spark tests skipping with optional dependencies absent, which is why a passed-only calculation would look 2 short.

A final cascade of four fixes then landed below this branch, taking it from 10088 to **10098**: **+1** walker probe of mine (below) and **+9** from those fixes.

### Windows CI

`test_budget.py::test_wall_time_clock_raises_after_deadline` fails on Windows for sibling PRs in this chain. It is pre-existing upstream code, present at `2965ec10` and absent from every diff in the chain — a `time.monotonic()` granularity issue, not ours.

`pre-commit run --files <the six changed files>`: **all hooks pass** — trailing whitespace, end-of-file, large files, case conflicts, merge conflicts, debug statements, mixed line ending, ruff check, ruff format, bandit, secret detection.

## Acceptance criteria

| Criterion | Proven by |
| --- | --- |
| Reports contain counts, durations, failures, fingerprints | `test_report_carries_counts_durations_failures_and_fingerprints` |
| No raw document text | `test_serialized_report_is_free_of_planted_phi` — scans `json.dumps(...)` of the whole payload, not field by field |
| Operators can resume from the CLI | `test_resume_recomputes_only_the_failed_shards` — attempts land as `[1, 1, 1, 2]`, so only the repaired shard re-ran |
| Operators can resume programmatically | `test_preserved_digest_on_a_requeued_shard_is_not_an_inconsistency` |
| Scanner fails if text reaches manifest or report | `test_start_output_and_manifest_contain_no_phi` — scans manifest bytes, report-file bytes and `caplog` records, and plants a known leak to prove the assertion is not vacuous |

The privacy guard is proven able to fail in three ways: PHI as a **mapping key** with a benign integer value (the test first shows a values-only scan finds nothing, so the key path is genuinely exercised), PHI as a value under a benign key, and an exception message smuggled into a field. Error messages are asserted never to quote the offending value, since `CliError` renders text verbatim and would otherwise reintroduce the leak.


maziyarpanahi:

PR #2031, opened 2026-08-02
feat: add Ray and Spark adapters for distributed batch de-identification

### LINKED ISSUE #1072: Add Ray and Spark adapters for distributed batch de-identification

PouyanJay:
Closes #1072

## Relationship to the rest of the OM-632 batch chain — read this first

This branch **contains the commits of #1069, #1071 and #1070**, so it shows 8 commits while closing a single issue. That is stacking, not scope creep.

| commit | issue | owner |
| --- | --- | --- |
| `ceff36de` | #1069 durable run manifest | sibling PR |
| `aa7e95f3` | #1069 portable unreadable-manifest test | sibling PR |
| `c1414901` | #1071 local executor + idempotent shard writes | sibling PR |
| `d4d3eaf3` | #1071 stop relying on inode identity | sibling PR |
| `2083014e` | #1071 keep visited nodes alive in the driver-only guard | sibling PR |
| `97f946a7` | #1070 resumable runs + straggler recovery | sibling PR |
| `f7782006` | **this PR** — `ray` extra, lockfile, backend fakes | |
| `cd5e5443` | **this PR** — Ray and Spark adapters | |

Only the last two are mine to review. Intended merge order is **#1069 → #1071 → #1070 → this PR**, and #1073 (PHI-free reports and CLI resume) stacks above this one.

**Rebase promise:** as each parent merges, I will rebase this branch onto the updated `master` and force-update it, so by the time this is considered the diff will contain only my two commits. It has already been replayed cleanly through four chain re-stackings; the current base is `2965ec10`.

## Summary

Adds optional Ray and Spark executor adapters behind the distributed batch `ShardExecutor` protocol.

- `RayShardExecutor` and `SparkShardExecutor` import their backend lazily inside `execute`, so `import openmed.processing` requires neither Ray nor PySpark.
- The same sharding and resumability conformance suite runs across local, Ray and Spark.
- Adapter failures stay isolated from the core local-first package when the optional dependencies are absent.

The worker side is the existing module-level `execute_shard_task` taking a plain `ShardTask`. Nothing else crosses to a worker, so a manifest store cannot reach one by construction; the shared driver-only-state guard runs eagerly before the first submission as defence in depth rather than as the primary control. Neither adapter reads driver-owned state from a module-level global, which `import` would re-create separately in every remote worker — a divergence no payload inspection can catch.

## Three findings worth stating plainly

**1. The Ray error unwrap was inert on a real cluster, and the test could only confirm the belief it encoded.**

The first version of `_remote_error_type` walked `__cause__` / `__context__`. Real Ray sets neither. `RayTaskError.as_instanceof_cause` builds a dual class that stores the original exception in a plain `.cause` **attribute**; `__cause__` is a `BaseException` getset descriptor, so it resolves to `None` rather than falling through the class's `__getattr__`, and the raise site is not inside an `except` block so nothing chains implicitly. The result was that every Ray remote failure recorded `UnknownError` — and since exception messages are forbidden in the manifest, `error_type` is the only failure signal, so a whole run's failures became indistinguishable.

The fix reads `getattr(exc, "cause", None)` first, keeping `__cause__` / `__context__` as fallbacks for wrappers that do chain normally. **The root fix was in the test fake**, not the assertion: `make_ray_task_error()` now reproduces `as_instanceof_cause` instead of hand-assigning `__cause__`. The old fake encoded a belief about Ray rather than Ray's behaviour, so the test could only ever confirm that belief. The test now asserts `exc.__cause__ is None` and `_safe_error_type(exc) == "UnknownError"` *before* asserting the unwrap, so it fails if either premise stops holding.

Spark needs no unwrapping: `Py4JJavaError` and `SparkException` are already valid identifiers and survive the shared sanitizer intact.

**2. `ensure_available()` does not prevent a burned manifest attempt, and an earlier rationale claiming it did was wrong.**

`run_shard_plan` marks every shard `RUNNING` and increments `attempts` *before* it calls the executor at all. Resolving the backend eagerly inside `execute` therefore cannot protect the manifest; it only makes the failure land at a fixed point with nothing submitted, instead of part-way through iteration. `ensure_available()` protects the manifest only when a caller invokes it *before* `run_shard_plan`.

This is now pinned by a pair of tests rather than a docstring: `test_missing_backend_still_costs_a_manifest_attempt` asserts `attempts == [1, 1, 1]` with all shards `RUNNING`, and `test_ensure_available_fails_before_any_attempt_is_burned` asserts `[0, 0, 0]`. For Ray, `ensure_available()` also covers runtime initialisation, since an unreachable cluster is a likelier failure than a missing import.

**3. My own cause-chain walker silently defeated the feature it exists to serve.**

`_remote_error_type` memoized visited nodes in a `seen: set[int]` keyed on `id()` — the same pattern as the driver-only-state guard bug fixed in `2083014e` below this commit. An id is unique only among *live* objects. A wrapper whose `cause` is a computed property returns a fresh exception per read, so each hop is freed as the walk advances, CPython reissues the address, and the next hop inherits a memoized id and is skipped. The walk stops early and reports `UnknownError`:

| chain depth | before | after |
| --- | --- | --- |
| 1 | `ValueError` | `ValueError` |
| 5 | **`UnknownError`** | `ValueError` |
| 50 | **`UnknownError`** | `ValueError` |
| 300 | **`UnknownError`** | `ValueError` |

`UnknownError` is precisely the diagnostics collapse this function was written to prevent, so the bug quietly reverted the feature to the behaviour it was fixing — while every existing test still passed.

I found this by **probing, not by inspection**: constructing a wrapper whose `cause` is a computed property and walking chains of increasing depth. Reading the code would not have surfaced it, because the code is correct in principle and only wrong about object lifetimes. The fix reuses `2083014e`'s mechanism — hold every visited node alive for the duration of the walk — so the codebase has one discipline for this hazard rather than two. Cycle safety is unaffected, and a parameterized regression test covers depths 1/5/50/300.

**Scope, honestly:** real Ray stores `.cause` as a plain attribute, so the parent holds the chain alive and this would not trigger on a real cluster today. But a wrapper with a computed `cause` is not exotic, and the failure mode is silent — no exception, no log, just a less useful manifest.

## Two narrower claims, corrected

- **Serializer strictness.** PySpark's `parallelize` ships elements with `CPickleSerializer`, which *is* stdlib pickle, so the handler-picklability gate is exactly as strict for Spark as for `ProcessPoolExecutor`. Only Ray is more permissive, using `cloudpickle`. An earlier comment claimed the fakes were stricter than all the real backends; they are stricter than one.
- **Retries and duplicate execution.** Neither adapter configures retry policy. Ray retries a task on worker death by default and Spark may run speculative duplicates, so a shard can execute more than once. This is safe because the write path is idempotent — temp file plus atomic replace, with a re-executed shard compared against its earlier digest before its output is replaced. The uncovered case is a handler that is not deterministic; that is reported as a digest mismatch with existing output left untouched, rather than silently overwritten. Documented where an operator will see it, not only in a module docstring.

## Files, and why each is here

The issue listed no files. Everything below is therefore a deviation by definition; each is listed with what forces it.

**Core**

| file | why |
| --- | --- |
| `openmed/processing/ray_executor.py` | `RayShardExecutor` |
| `openmed/processing/spark_executor.py` | `SparkShardExecutor` |
| `openmed/processing/__init__.py` | exports both. Additive only — `check_public_api.py` fails on `change.breaking`, so additions pass `test_public_api_compat.py` without a baseline regen |

**Tests**

| file | why |
| --- | --- |
| `tests/unit/processing/test_distributed_executor_conformance.py` | the cross-backend suite the issue asks for |
| `tests/unit/processing/test_distributed_backend_adapters.py` | adapter-specific contracts: lazy import, remote-error naming, submission bounding |
| `tests/unit/processing/fixtures/distributed_backend_fakes.py`, `fixtures/__init__.py` | Ray/Spark fakes. Mirrors the existing `tests/unit/clinical/fixtures/dhis2/` helper-package convention |
| `tests/unit/processing/test_distributed_backend_fakes.py` | the fakes stand in for absent backends, so they are themselves tested |

**Forced by gates**

| file | gating test |
| --- | --- |
| `pyproject.toml` | new `ray` extra |
| `uv.lock` | `.github/workflows/ci.yml:170` runs `uv lock --check`; must land in the same commit as the pyproject change |
| `scripts/release/check_license_policy.py` | `tests/unit/release/test_license_policy.py::test_current_non_dev_dependencies_are_permissive`. **Non-obvious trap:** PyPI ships Ray's license as the string `"Apache 2.0"` — with a space — which matches no entry in `ALLOWED_LICENSE_MARKERS` (`APACHE-2.0`, `APACHE SOFTWARE LICENSE`), and `contains_marker` normalizes `_`→`-` but not spaces. Without a reviewed-licenses entry an unreviewed `ray` resolves to not-allowed and fails the policy test. Any future contributor adding an extra whose upstream license string is punctuated differently will hit this |
| `tests/unit/interop/test_core_does_not_import_adapters.py` | `"ray"` added to `OPTIONAL_ADAPTER_MODULE_PREFIXES`, mirroring the `pyspark` entry added by the Spark UDF PR |

**Docs** — appended to pages already in the nav; no new page, so no `mkdocs.yml` and no `docs/brand/system/publication.yml` change was needed (`test_docs_publication.py` asserts index parity over the *set* of markdown files, which appending does not alter).

| file | why |
| --- | --- |
| `docs/batch-processing.md` | `## Distributed shard execution`. Already in the nav and already owns checkpoints/resume. Note `docs/spark-deidentification.md` was the wrong home — it documents the `pandas_udf` column-redaction feature, a different thing |
| `docs/troubleshooting.md` | `ray` / `spark` row in the extras table, naming both classes so anyone who sees them in `__all__` can find the install line |

## Optional dependency and licensing

New `ray = ["ray>=2.30,<3"]` extra. Core Ray only — the executor uses `ray.remote`/`ray.get`, not `ray[data]`. The existing `spark` extra is reused unchanged.

`uv lock` regenerated and committed in the same commit as the `pyproject.toml` change:

- added `ray 2.56.1` and transitive `msgpack 1.2.1`
- **0 removals, 0 version changes** on any existing package (verified programmatically over every `[[package]]` name/version pair, 454 → 456)
- `uv lock --check` exit 0
- `"ray": "Apache-2.0"` added to `REVIEWED_LICENSES`; audit reports `ray → Apache-2.0, allowed=True`, zero disallowed overall

## What is proven here, and what is not

**Proven offline, all three backends:** protocol conformance; identical output digests across local, Ray and Spark; result attribution by `shard_id` rather than completion order (Ray driven with an explicit out-of-order completion sequence, and with a bounded `max_in_flight` window); PHI-free failure reporting; empty shards settling as `COMPLETED`; single-shard plans; `expected_digest` preserved across a resume, with a non-determinism case asserting existing output survives; rejection of a captured manifest store before a single task is submitted; worker-death and job-failure paths driven through the adapters' infrastructure-failure branches.

**Not proven — needs a live cluster.** The real-backend tests are **skipped, not passing**. CI runs `uv sync --frozen --extra dev` only (`ci.yml:134,198,248,291,339`), so neither `ray` nor `pyspark` is ever installed there and these will skip in CI exactly as the existing `test_spark_udf.py` and `test_spark_streaming.py` already do. To exercise them locally:

```bash
uv sync --extra dev --extra ray     # then: python -m pytest tests/unit/processing -q
uv sync --extra dev --extra spark
```

Nothing about real `cloudpickle` serialization, per-worker model loading, atomic rename on a shared filesystem, executor loss and re-execution, or Ray object-store spilling has been verified. Related: the class name `RayTaskError(ValueError)` is a **synthetic reconstruction** modelled on Ray's source, since Ray is not installed in this environment — the unwrapping mechanism is proven against that shape, the exact rendered upstream name is not.

Adapter coverage: `ray_executor.py` **100%**, `spark_executor.py` **96%** — the two uncovered lines are `return module.SparkSession` and `return active`, reachable only with PySpark actually installed.

## Data provenance

All fixture data is synthetic and generated algorithmically — documents are built as `f"doc-{index:05d}"` / `f"synthetic clinical note {index}"`, shard fingerprints and digests are derived from those values. No real patient data, no copied records, no network access in any test.

PHI containment is asserted, not assumed: the conformance suite raises a handler failure whose exception message interpolates `document_hashes`, then scans the serialized run result to assert no document hash appears. Workers return exception *type* names only, never messages, since messages can quote document text.

## Verification

Full suite at base `2965ec10`:

| --- | --- | --- | --- | --- |
| baseline `2965ec10` | 31 | 9816 | 86 | 1 |
| this branch | 31 | 10049 | 88 | 1 |

**FAILED sets are byte-identical** (31 = 31, sorted diff empty). The 31 pre-existing failures are unrelated — zh-segmentation/normalize (jieba/opencc), sigstore, MCP entrypoint, brand-system, and an API-surface migration-guide test.

The +233 passed decomposes exactly: **139** from the chain commits' own test files (`test_batch_run_manifest.py`, `test_shard_executor.py`, `test_batch_resume.py`, including the regression tests added by `aa7e95f3`/`d4d3eaf3`/`2083014e`) plus **94** from mine. My three files collect 96: **94 passed + 2 skipped**, the 2 skips being the real-backend tests, which also accounts for 86 → 88 skipped.

**Guard verification with a negative control.** The driver-only-state guard these adapters depend on was fixed in `2083014e`. Rather than assume the fix carried, I drove the memo-poisoning payload (N `functools.partial` handlers, the manifest store captured in the last one) through all three backends at N = 20 / 100 / 200 — rejected in all nine cases. I then ran the identical payload against the **pre-fix** guard at `d4d3eaf3` and confirmed it missed at all three sizes. A probe that cannot fail proves nothing, so the negative control is the part that makes the positive result meaningful.

**Iteration-order and hash-seed stability.** A fake that models completion ordering and a bounded in-flight window is exactly where an iteration-order dependence would hide, so this was measured rather than argued: **5 `PYTHONHASHSEED` values** (0, 1, 42, 12345, 99999) × **3 randomised test orderings** — `94 passed, 2 skipped` on every one of the eight runs. The adapter's `seen` set is membership-only and never iterated; `in_flight` is keyed by object reference and results are attributed by `shard_id`, never by position.

**Windows CI note, not from this branch.** `tests/unit/core/test_budget.py::test_wall_time_clock_raises_after_deadline` fails on Windows for sibling PRs in this stack. It is pre-existing upstream code — present at `2965ec10` and absent from every commit in this diff — caused by `time.monotonic()` granularity on Windows. Not introduced here.

`pre-commit run --files` over all 14 files in my two commits: **all hooks pass**, no reverts required — nothing under `docs/` beyond my two appended sections and nothing under `tests/fixtures/clinical/`, so the repo's known legacy end-of-file churn was not triggered.


maziyarpanahi:

PR #1558, opened 2026-07-13
Add pluggable Chinese word segmentation

maziyarpanahi:
# Pull Request

## Description

Closes #1555.

Adds the independently mergeable Chinese segmentation foundation from parent epic #1465: exact-offset word segmentation, a synthetic clinical dictionary, configurable optional backends, and catalog-backed Chinese routing. The implementation is offline-first and never downloads optional model assets implicitly.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition/improvement

## Changes Made

- Added a common segmentation protocol, exact-offset validation, a lightweight default backend, and lazy optional adapters.
- Added a seeded synthetic medical dictionary, user-dictionary loading, backend configuration, and offline model-asset safeguards.
- Registered Chinese through the language-pack catalog, Han routing, release manifests, service schemas, clients, synthetic data, generated tool schemas, and public documentation.
- Kept the Chinese default model explicit as an API-compatible placeholder rather than claiming dedicated trained weights.
- Added a 200-sentence boundary-quality gate, 1,000-string round-trip coverage, and focused adapter/configuration tests.

## Testing

- [x] Added tests that prove the feature works
- [x] New and existing unit tests pass locally
- [x] Tested different backends and inputs

Validation completed:

- .venv/bin/python -m pytest tests/ -q — 5,327 passed, 52 skipped
- Focused segmentation, routing, registry, release-gate, and schema tests — 798 passed, 5 skipped
- make format
- make lint
- make format-check
- make docs-build
- make build
- uv lock --check
- .venv/bin/python scripts/release/check_license_policy.py

## Documentation

- [x] Updated user-facing documentation
- [x] Added docstrings to public functions and classes
- [ ] Updated CHANGELOG.md

## Code Quality

- [x] Ran make format, make lint, and make format-check
- [ ] For Swift/OpenMedKit changes, ran make format-swift and make lint-swift
- [x] Performed a self-review
- [x] Added comments around offset alignment and optional model handling
- [x] The change generates no project warnings

## Dependencies

- [ ] No new dependencies
- [x] Added justified dependencies: jieba is the lightweight core segmenter; pkuseg and HanLP remain opt-in extras under permissive licenses.

## Checklist

- [x] Read the contributing guidelines
[...]

Issue #1557, opened 2026-07-13
Document Chinese segmentation dictionaries and model operations


maziyarpanahi:
Parent epic: #1465

## Size
M

## Scope
- Add a dedicated operations guide for locally provisioned pkuseg and HanLP model assets after their real-model conformance work lands.
- Document tested model-version compatibility, cache and storage layout, deployment matrices, and upgrade/rollback procedures.
- Add governance guidance for merging institutional dictionaries, recording provenance, validating entries, and running segmentation regressions.
- Add runnable optional-backend examples using the conformance fixtures from #1556.

## Acceptance criteria
- Each optional backend has a tested end-to-end local-model example.
- Supported model versions and deployment requirements are explicit.
- Dictionary governance and regression-check procedures are documented.
- The strict documentation build passes.

Blocked by #1556. The base dictionary format and offline guarantees ship with the first child, #1555; this follow-up covers real-model operations.

PR #2034, opened 2026-08-02
test: validate optional Chinese segmenters with installed domain models

### LINKED ISSUE #1556: Validate optional Chinese segmenters with installed domain models
Parent epic: #1465

## Size
M

## Scope
- Add opt-in conformance tests for installed pkuseg medicine and HanLP models.
- Compare backend token offsets and dictionary overrides against the common protocol.
- Record performance and boundary-quality evidence without downloading or bundling model files.
- Keep all default tests offline-friendly.

## Acceptance criteria
- Each installed backend passes the shared exact-offset conformance suite.
- External model tests are explicitly marked and skipped when assets are absent.
- No licensed corpora or model weights are committed.

Blocked by the first child: #1555.

PouyanJay:
Closes #1556

Adds a shared exact-offset conformance suite for the pluggable Chinese
segmentation backends introduced in #1558, and enrolls really installed
pkuseg `medicine` and HanLP models through it. No model weights, licensed
corpora, or new dependencies are added.

## Scope

| Issue scope item | Where |
|---|---|
| Opt-in conformance tests for installed pkuseg medicine and HanLP models | `test_installed_backend_passes_shared_conformance_suite`, parameterized over jieba/pkuseg/hanlp |
| Compare backend token offsets and dictionary overrides against the common protocol | `run_segmenter_conformance()` + `test_installed_backends_agree_on_dictionary_override_spans` |
| Record performance and boundary-quality evidence without downloading model files | `SegmentationConformanceReport.to_evidence()`, emitted via `record_property` |
| Keep all default tests offline-friendly | Layer A: 28 tests, no optional dependency required |

## Acceptance criteria

- **Each installed backend passes the shared exact-offset conformance suite** —
  `test_installed_backend_passes_shared_conformance_suite[jieba|pkuseg|hanlp]`.
- **External model tests are explicitly marked and skipped when assets are
  absent** — all Layer B tests are `@pytest.mark.integration` and skip via
  `pytest.importorskip` plus an asset probe (pkuseg's `medicine` model
  directory; `OPENMED_HANLP_MODEL_PATH` for HanLP).
- **No licensed corpora or model weights are committed** — enforced by the
  existing `tests/unit/release/test_license_policy.py` package-data and wheel
  audits, which this PR does not modify.

## Design: two layers

**Layer A (always runs, no optional dependency).** The harness is exercised by
an in-repo reference segmenter plus deliberately broken stubs. Each of the
seven declared checks — `protocol, alignment, offsets, overlap, ordering,
coverage, dictionary` — has a stub that triggers it, and the assertions are
exact `checks_triggered` set equality rather than "not empty", so a stub that
fails for the wrong reason is caught.

**Layer B (opt-in).** Marked `integration`; skips with an explicit reason
naming the extra and its license when a backend or a user-supplied model is
absent. OpenMed never downloads model files.

### Why the thresholds are where they are

This matters for reviewing the suite as anything other than over-engineering.
The 200-case corpus on its own **passed two grossly broken segmenters**: one
that emits every dictionary term correctly but merges all other runs into a
single token (boundary F1 `0.972`), and one that additionally splits every
non-dictionary run into single characters (`0.807`). Both satisfied the span
protocol and every dictionary override. That is why the checks and the floor
exist in their current form.

`SEGMENTATION_BOUNDARY_F1_FLOOR = 0.90` — matching the default-backend gate
already enforced in `test_zh_segmentation.py` — rejects the character-level
case. It does **not** reject the merge-into-blobs case at `0.972`. Rather than
hide that, `test_boundary_f1_floor_catches_a_character_level_segmenter`
asserts both facts, so the scope limit is executable and the next person to
close the gap starts from a test that documents where the line sits. Raising
the floor further would exceed the shipped default backend's own gate and
would likely fail legitimate user-supplied HanLP models.

### The two-layer threshold split is deliberate

Same metric, two purposes:

| | threshold | why |
|---|---|---|
| in-repo reference segmenter | exactly `1.0` | fixed implementation over a checksum-pinned corpus; anything less means the harness itself broke |
| installed backends | `>= 0.90` | output legitimately varies by backend build and user-supplied model |

`chars_per_second` is asserted only as `> 0.0`. That is a **liveness check that
the timer ran — it judges nothing about the number**, and it is the one
genuinely hardware-dependent metric. It is not a throughput standard.

## Files

The issue lists no files, so each is noted with the gate that forces it.

| File | Why |
|---|---|
| `tests/unit/processing/test_zh_segmentation_conformance.py` (new) | the suite |
| `openmed/processing/zh_segmentation.py` | the harness must be library code, not test-local, so it is importable and its own behaviour is testable offline |
| `openmed/processing/__init__.py` | package re-exports; pinned by `test_every_conformance_name_resolves_from_the_package_namespace` |
| `tests/fixtures/processing/zh_segmentation_conformance.json` (new) | corpus; pinned by `test_conformance_corpus_is_synthetic_and_reproducible` (sha256) and `test_fixture_dictionary_terms_match_the_shipped_medical_dictionary` |

`SEGMENTATION_BOUNDARY_F1_FLOOR` is re-exported from `openmed.processing`
alongside the other six conformance names. It was initially the only one
present in the submodule's `__all__` and absent from the package — a named
threshold callers cannot import defeats its purpose. The uniformity test pins
all seven **as a set**, which is the shape that catches the next such omission
regardless of which name it hits.

### Deliberate omissions

- **`scripts/release/public_api_baseline.json` is not regenerated.** Running
  `check_public_api.py --update` produces 390 insertions / 10 deletions, of
  which only 5 lines are from this PR; the rest is unrelated already-merged
  work (`CLINICAL_DECISION_SUPPORT_*`, `GuardedSuggestion`,
  `SUBJECT_SURROGATE_LABEL`, and others) that never refreshed the baseline.
  Sweeping that into a Chinese-segmenter change would misattribute a large
  unrelated diff. Additions are non-breaking (`unannounced_breaks` filters on
  `change.breaking`), and the gate is green without it. **Suggest a separate
  maintenance PR to refresh the baseline.**
- **pkuseg and HanLP are not moved into the `zh` extra.** They stay in
  `zh-pkuseg` / `zh-hanlp`. `tests/unit/release/test_license_policy.py`
  asserts exact dict equality over the `zh` and `indic` groups, so moving them
  would break it — and the point of separate extras is that these backends are
  optional and model-dependent.
- **No dependency added, so no `uv.lock` change.** Both extras and both
  license entries already existed.

### One note on a skip

The new suite `importorskip`s **jieba**, which is a core dependency. This is
not concealment: jieba is always installed in CI, so the suite never skips
there, and the four existing un-skipped jieba tests in `test_zh_segmentation.py`
remain the packaging canary. The skip only keeps the new suite usable in
development environments without the optional language stack.

## Data provenance

All data is **synthetic and algorithmically generated**. The 200-case corpus
is composed by Cartesian product of names and conditions taken from the
already-shipped synthetic dictionary `openmed/processing/data/zh_medical_terms.txt`
across four templates, pinned by sha256 and verified by recomputation in
`test_conformance_corpus_is_synthetic_and_reproducible`. No real patient text,
no PHI, no external or licensed corpus, and no model files are included.

## Verification

Canonical baseline at `2965ec10`: **31 failed / 9816 passed**.
This branch: **31 failed / 9844 passed / 90 skipped / 1 xfailed**.

- Sorted FAILED sets are **identical** to the canonical baseline — no
  regression. The 31 pre-existing failures are environment-driven (absent
  optional dependencies) and unrelated to this change.
- Passed delta reconciles exactly: `9816 + 28 = 9844`, the 28 new Layer A
  tests. The 4 additional skips are the Layer B tests correctly skipping
  without pkuseg/HanLP installed.
- `pre-commit run --files <the four files>`: all hooks pass.

Rebased onto `2965ec10` (the OM-801 input-normalization gateway) and
re-verified. That change adds `normalize_text()`, so I checked it cannot
destabilise the exact-`1.0` reference threshold: it performs no Unicode
normalization (no NFC/NFKC, no width folding, only whitespace stripping and
size limits), and no import path reaches it from `zh_segmentation.py` or
`tokenization.py`. Empirically it alters 0 of the 200 corpus texts and
preserves the full-width probe byte-for-byte.

## Relationship to other work

#1557 (document Chinese segmentation dictionaries and model operations) stacks
on this branch and will rebase once this merges. It documents the `integration`
marker, the fixture path, `OPENMED_HANLP_MODEL_PATH`, and the threshold table
above.


maziyarpanahi:

PR #1562, opened 2026-07-13
Add Chinese clinical NER evaluation foundation

maziyarpanahi:
## Description

Adds the independently mergeable Chinese clinical NER evaluation foundation from child issue #1559. The parent epic remains open for the optional-model and broader CBLUE follow-up work in #1560 and #1561.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [x] Test addition/improvement

## Changes Made

- Map all nine CMeEE categories through Chinese-aware canonical label normalization, including normalized long-form aliases, with no `OTHER` fallback for the seven core clinical types.
- Reuse the explicit-path, license-boundary-enforcing CMeEE loader and keep CMeEE, CBLUE, eHealth records, and related model weights outside the repository.
- Add a bundled two-record synthetic CMeEE-shaped fixture for deterministic offline CI.
- Add the `chinese-clinical-ner` suite with exact overall and per-label precision/recall plus a zero-tolerance PHI-token leakage gate.
- Keep leakage evidence PHI-safe: offsets, labels, fixture IDs, and SHA-256 hashes only.
- Preserve current master's Chinese model contract: `zh` uses the documented multilingual routing placeholder and is not advertised as a manifest-backed dedicated Chinese checkpoint.
- Add Chinese-script coverage to deterministic synthetic locale tests.

## Testing

- [x] Full repository suite: `5785 passed, 47 skipped`
- [x] Focused eval, multilingual loader, label, registry, and synthetic-locale suite: `302 passed, 5 skipped`
- [x] Synthetic safe and deliberately leaky runs both exercised.
- [x] `make format`, `make lint`, and `make format-check`
- [x] Scoped pre-commit hooks, including Bandit and secret detection
- [x] `.venv/bin/python scripts/release/check_license_policy.py`
- [x] `make docs-build`
- [x] `/usr/local/bin/python3 -m build`

## Documentation

- [x] Documented the suite, zero-leakage gate, current routing-placeholder boundary, and external licensed-asset requirements.
- [x] New public functions and classes include docstrings.
- [ ] CHANGELOG updated; not required for this issue-scoped foundation PR.

## Code Quality

- [x] Rebased onto current `master` and resolved overlapping language/eval surfaces without stale broad formatting or documentation churn.
- [x] No new dependencies.
- [x] No restricted corpus records or model weights are committed.

## Related Issues

Closes #1559
Related to #1473
[...]

Issue #1561, opened 2026-07-13
Expand CBLUE evaluation coverage and Chinese model operations guidance


maziyarpanahi:
Parent epic: #1473

## Size
M

## Scope
- Extend the user-supplied benchmark interface to additional CBLUE task shapes that are relevant to Chinese clinical entity evaluation.
- Add synthetic smoke fixtures and provenance checks for each supported task without implementing CMeIE relation decoding.
- Document tested local-asset layouts, model compatibility, evaluation commands, result interpretation, and upgrade or rollback procedures.
- Add model-card evidence guidance for multilingual fallback and dedicated local models.

## Acceptance criteria
- Each added task shape has an offline synthetic smoke test and an explicit external-path boundary.
- CMeIE relation decoding remains out of scope.
- Operations and model-card guidance identifies license, provenance, storage, and leakage requirements.
- The strict documentation build and full test suite pass.

Blocked by optional-model validation from #1473.

Issue #1559, opened 2026-07-13
Add the Chinese clinical NER evaluation foundation and leakage gates


maziyarpanahi:
Parent epic: #1473

## Size
L

## Scope
- Register `zh` across the supported-language, display-name, model-prefix, default-model, model-registry, and locale surfaces, using the existing multilingual privacy filter as the offline-safe fallback.
- Add language-aware CMeEE label normalization for all nine source categories, with explicit canonical mappings for the core clinical types and department.
- Build a dedicated Chinese clinical NER suite on the existing explicit-path CMeEE loader, including a bundled synthetic fixture, per-label precision and recall, and PHI-token leakage reporting.
- Enforce a zero-leakage gate for synthetic identifiers and document that CMeEE, CBLUE, eHealth data, and model weights remain user-supplied external assets.

## Acceptance criteria
- Chinese is reported as supported and `get_pii_models_by_language("zh")` returns the multilingual fallback.
- All nine CMeEE categories normalize to canonical labels; the core clinical types never fall back to `OTHER`.
- The synthetic suite runs offline, reports per-label metrics, and fails when any injected identifier survives.
- Missing or repository-internal real CMeEE paths fail with clear license-boundary guidance.
- No restricted corpus records or model weights are committed, and the full test suite passes.

This is the first independently mergeable child of #1473.

PR #2035, opened 2026-08-02
feat: add opt-in Chinese clinical NER local-asset conformance

### LINKED ISSUE #1560: Validate optional Chinese clinical NER models with local assets
Parent epic: #1473

## Size
M

## Scope
- Add opt-in conformance adapters for Chinese clinical NER model weights provisioned by users outside the repository.
- Compare model span offsets and canonical-label outputs against the Chinese clinical NER suite.
- Record per-label quality, latency, and leakage evidence without downloading or bundling restricted weights or corpora.
- Keep all default tests offline-friendly and skip optional model checks when local assets are absent.

## Acceptance criteria
- Each configured local model passes the common exact-offset and canonical-label contract.
- External-model tests are explicitly marked and skip with actionable configuration guidance when assets are absent.
- Evaluation reports include per-label metrics and PHI leakage without raw identifiers.
- No licensed corpus records or model weights are committed.

Blocked by the first child of #1473.

PouyanJay:
## Summary

Adds an opt-in conformance path for Chinese clinical NER checkpoints that users provision **outside** this repository. `OPENMED_ZH_CLINICAL_NER_MODEL_DIR` mirrors the existing `OPENMED_CMEEE_PATH` boundary: nothing is downloaded, vendored, or auto-discovered.

Scope items from the issue:

- **Opt-in conformance adapters for user-provisioned weights** — `ChineseClinicalNerAdapter` protocol, `resolve_chinese_local_model_asset`, `chinese_local_model_asset_skip_reason`.
- **Compare span offsets and canonical-label outputs against the suite** — `run_chinese_clinical_ner_conformance` enforces the contract rather than warning about it.
- **Per-label quality, latency, and leakage evidence without raw identifiers** — latency percentiles added on top of the suite's existing `per_label` and `phi_token_leakage`; all evidence is hashed or offset-only.
- **Default tests stay offline; optional checks skip when assets are absent** — the opt-in test skips with the environment variable named in the reason.

The foundation this builds on (the `chinese-clinical-ner` suite, the CMeEE loader, and the label mapping) landed in #1562 / #1559; this PR adds only the local-asset layer.

Closes #1560

## What "conformance" does and does not prove

Stated plainly because the wording matters:

- Passing proves the conformance **contract** and the **skip semantics**. It does not prove the measured quality of any checkpoint — quality numbers only exist for a user's own opt-in run against their own licensed corpus. This is unavoidable: the repository is offline-first and bundles no weights.
- The contract is that every predicted span is well formed against the fixture text — in bounds, non-inverted, non-empty, canonically labelled, and exactly aligned to the characters it claims. Violations raise `ChineseClinicalNerContractError` with surface-free evidence (offsets, label, problem codes — never span text).

Two behaviours are **intentionally legal**, and a reviewer may reasonably propose changing them, so the reasoning is recorded here:

- **Overlapping and nested spans stay legal.** CMeEE data is nested; the repository's own `tests/unit/eval/fixtures/license_aware_ner/zh_standin.json` carries spans 0–3 and 0–5 on a single record. Rejecting overlaps would reject the suite's intended input.
- **A +1-shifted span is a well-formed wrong answer**, not a contract breach. `exact_span_f1` already reports it as 0.000. Treating it as invalid would mean "predictions must equal gold", which is the metric's job, not the contract's.

## A category error in the opt-in test, and its fix

The opt-in integration test as first written **could not pass for any realistic CMeEE-trained checkpoint**, and because it is skip-guarded, CI could never have noticed. The cause is a category error rather than a coding slip:

The suite's `gate.passed` is driven by a zero-tolerance PHI-leakage check, which judges **de-identification** behaviour. But the artifact `OPENMED_ZH_CLINICAL_NER_MODEL_DIR` exists to point at is a clinical **entity** model. A CMeEE-trained checkpoint emits `bod`/`dis`/`sym`/`dru`/`equ`/`pro`/`ite`/`mic`/`dep` and no PHI types at all, so it never predicts the synthetic identifiers, the redactor never masks them, and the run raised `ChineseClinicalNerLeakageError` before reaching a single assertion. A user would have set the variable and received a stack trace instead of a scorecard.

The obvious fix has a trap worth recording: filtering to the clinical-only record instead trips `total_tokens == 0` → `no_synthetic_phi_tokens`, so the gate fails from the other direction. **Neither bundled record works alone.**

Resolution: run both records with `fail_on_leakage=False`, assert what a clinical entity model can actually satisfy (contract holds, `per_label` populated and a subset of `CANONICAL_LABELS`, latency count, hash-only leakage findings), and **deliberately do not assert `gate.passed`**. Leakage is read as evidence, not as a verdict on a model that was never a de-identifier. Both the module docstring and the test docstring say so.

## The no-weights guard deliberately exceeds "no weights are committed"

`test_repo_ships_no_zh_model_weights` scans tracked files **and untracked files that are not git-ignored**. This is intentional and stricter than the issue's literal wording.

Rationale: the accident this guard exists to prevent **begins as an untracked file**. By the time weights are committed the guard has already failed at its job. A checkpoint parked in the working tree is one `git add -A` away from the thing being guarded against, so a red test at that moment is the correct outcome rather than noise. The docstring records this too, but it is stated here so it does not have to be reverse-engineered.

False positives were removed rather than tolerated: the guard is scoped to weight-artifact **filenames** (Hugging Face conventions plus unambiguous checkpoint suffixes), so an unrelated `testdata.npz` cannot trip it; it ignores symlinks, since "keep the checkpoint outside the repository" is the recommended workflow; and it skips cleanly when git is unavailable instead of erroring.

**Suggestion for the maintainer:** adding `*.safetensors` and `*.bin` to `.gitignore` would turn this strictness into a non-event for local checkout work — a remedy that does not require softening the guard. Not done in this PR, since `.gitignore` is outside the issue's scope.

## Files changed, including those the issue did not list

The issue named no files. Deviations from its letter, each with the gate that forces it:

| File | Why |
|---|---|
| `openmed/eval/suites/chinese_clinical_ner.py` | The adapter protocol, asset resolver, and conformance runner. |
| `tests/unit/eval/test_chinese_clinical_ner.py` | Suite tests live in one file, mirroring #1562's structure. |
| `openmed/eval/suites/__init__.py` | **Extra.** The package re-exports every suite symbol; omitting the new ones leaves the public surface inconsistent with the eight sibling suites in the same block. Pure insertion, no deletions. |
| `docs/eval-harness.md` | **Extra.** #1562 documented the suite here, and the strict docs build covers it. Pure insertion of one `###` subsection. |

Deliberately **not** touched, and each was verified to stay green:

- **`tests/unit/eval/test_metrics.py:39`** holds an exact-tuple gate `assert DEFAULT_SUITES == (...)`. No new suite id is registered — the work is a function on the existing `chinese-clinical-ner` suite — so this gate stays untouched. Registering an id would have forced an edit here plus a docs row for a suite that cannot run in CI.
- **`pyproject.toml`** — the existing `integration` marker is reused rather than adding a new one.
- **`models.jsonl`** — no registry change is required.

Naming note: the local-asset symbols are `Chinese`-qualified (`ChineseLocalModelAsset`, `resolve_chinese_local_model_asset`, `chinese_local_model_asset_skip_reason`) because `openmed.eval.suites` is shared by eight suites and every sibling in that block carries the same qualifier.

## Data provenance

All test data is **synthetic and algorithmically generated**. No real PHI, no real clinical Chinese text, no licensed corpus records, and no model weights are committed.

The existing `openmed/eval/fixtures/cmeee_zh_synthetic.jsonl` is reused **unchanged** — it already carries all nine CMeEE categories plus three synthetic PHI spans, so no new fixture data was needed. Local checkpoint directories in tests are built with `tmp_path` from empty files; no weights are ever read. Leakage findings carry only fixture ids, offsets, canonical labels, and SHA-256 hashes; asset evidence carries a path fingerprint plus whitelisted artifact descriptors, never the local path, the directory basename, or an operator-chosen filename.

## Verification

Rebased onto `2965ec10`; `git log 2965ec10..HEAD` shows exactly one commit.

- **This branch:** 31 failed, 9838 passed, 87 skipped, 1 xfailed
- **Canonical baseline at `2965ec10`:** 31 failed, 9816 passed
- **The FAILED sets are identical** — 31 entries, zero regressions. The 22 additional passes are this PR's new tests.
- `tests/unit/eval/test_chinese_clinical_ner.py`: 34 passed, 1 skipped (the skip is the opt-in check, skipping with the environment variable named in its reason).
- `pre-commit run --files <changed>`: all hooks pass — `ruff check`, `ruff format`, bandit, secret detection, EOF, whitespace, merge conflicts, debug statements.

The pre-existing 31 failures are environmental and unrelated to this change (OpenCC and jieba optional extras, a Sigstore bundle check, brand asset reproducibility, an MCP console entry point, and an API-surface migration-guide check); each was confirmed present on a clean baseline before this work.

Both guards in this PR were verified capable of failing rather than passing vacuously: planting a real `model.safetensors` makes the no-weights guard fail, and mutating the skip message to drop the environment variable name makes the guidance assertion fail. Both probes were removed and the tests re-verified green.

## Relationship to #1561

#1561 (expand CBLUE evaluation coverage and Chinese model operations guidance) **stacks on this branch** and rebases when this merges. It goes second by design, because this is the larger diff and replaying the smaller diff onto the larger is the cheaper direction.

Conflict surface: `openmed/eval/suites/chinese_clinical_ner.py` and `tests/unit/eval/test_chinese_clinical_ner.py` are the real overlap. `docs/eval-harness.md` and `openmed/eval/suites/__init__.py` were verified to auto-merge cleanly, because both changes here are pure insertions with zero deletions. Whoever lands second should **re-run `tests/unit/eval/test_chinese_clinical_ner.py` rather than trusting the merge**.

Happy to rebase on request if `master` moves before this is reviewed.


maziyarpanahi:

PR #2036, opened 2026-08-02
docs: add Chinese segmentation operations guide

### LINKED ISSUE #1557: Document Chinese segmentation dictionaries and model operations

PouyanJay:
Closes #1557.

## Relationship to #1556 — read first

**This branch contains #1556's commit (`df22ee55`), so it shows three commits while closing one issue.**

```
49140337 test: make Chinese segmentation doc assertions path-portable  <- this PR
92a1b02b docs: add Chinese segmentation operations guide               <- this PR
df22ee55 test: validate optional Chinese segmenters ...                <- #1556
2965ec10 Merge pull request #2023 ...                                  <- master
```

#1557 is explicitly *blocked by #1556*: this guide documents the conformance suite that #1556 ships, and its documentation test asserts the guide's claims directly against that suite's code. Stacking was the only way to write those assertions against something real rather than against a guess.

**#1556 must merge first.** Once it does, I will rebase this branch onto the updated `master` so it carries only its own commit, and confirm `git log master..HEAD` shows exactly one. If #1556 changes further before merging, I will rebase onto its new tip and re-verify every documented claim rather than assume the guide still holds — that has already happened four times during development, and each time the documentation test caught the drift for me.

## Summary

Adds `docs/chinese-segmentation-operations.md`, an operations guide for running the optional pkuseg and HanLP Chinese segmentation backends against locally provisioned model assets, mirroring the issue's four scope items:

- **Operations guide for locally provisioned pkuseg and HanLP assets** — declared version ranges, the configuration and environment-variable surface, pkuseg name-versus-path resolution and cache layout, and HanLP model provisioning.
- **Model-version compatibility, cache and storage layout, deployment matrices, upgrade/rollback** — a deployment matrix covering which backends need operator-provisioned assets, a "what fails when" table separating construction-time from first-`segment()` failures, and directory-swap upgrade/rollback procedures.
- **Dictionary governance** — provenance recording, merge order and per-backend precedence, the bounded-validation limits, and the rejection-error taxonomy.
- **Runnable optional-backend examples using #1556's conformance fixtures** — provisioning each backend, the shared corpus, running the harness, and what gates versus what is only recorded.

The guide states each operational rule **once**, as a table, and the accompanying test parses that table and executes it against the implementation.

The issue lists no file paths, so every file here is an addition worth accounting for. Four of the five are forced by gates rather than chosen:

| File | Why | Gating test |
| --- | --- | --- |
| `docs/chinese-segmentation-operations.md` | the guide itself | — |
| `mkdocs.yml` | nav entry | `tests/unit/test_docs_publication.py::test_publication_classifies_every_default_markdown_page_exactly_once` |
| `docs/brand/system/publication.yml` | publication classification, at the **same index** as the nav entry | same test — it compares `navigated` to the nav as **ordered lists**, so an entry at a different position fails even when both files contain it |
| `docs/anonymization.md` | cross-link from the existing Chinese segmentation section | `tests/unit/test_docs_publication.py::test_every_link_only_page_is_reachable_from_the_documentation_nav` |
| `tests/unit/processing/test_zh_segmentation_docs.py` | asserts the guide against the code | — (new) |

Any new page under `docs/` must be registered in **both** `mkdocs.yml` and `publication.yml`, at matching positions. That parity requirement is not discoverable from the issue.

Two further tests were added to that module after the Windows CI job surfaced a portability bug (see below):

| Test | Gate forcing it |
| --- | --- |
| `test_documentation_needles_are_posix_paths` | the Windows CI job — a doc-facing needle built with the local path separator cannot match a Markdown document |
| `test_guide_reads_as_utf8_on_every_platform` | the Windows CI job — the platform default encoding is not UTF-8 there |

Deliberately **not** included:

- **No CHANGELOG entry.** No test gates it, and it conflicts across concurrent branches.
- **No `.zh.md` translation.** `test_translation_groups_are_complete_without_false_fallback_routes` asserts the localized nav as an exact mapping; adding a localized variant of this page would fail it.
- **Not folded into `docs/on-device-segmenters.md`.** That page covers packaged on-device bundle resources, a different subject from runtime backends.

## Data provenance

All examples are **synthetic and algorithmically generated**. No real PHI, no real clinical text, no licensed corpora, no model weights. The names and conditions come from the shipped synthetic dictionary and #1556's generator-spec corpus (40 synthetic names x 5 conditions across 4 templates, checksum-pinned). The dictionary-validation example writes its own temporary file.

## Documentation tests execute the rules rather than search for them

The first version of the drift guard asserted that key phrases appeared somewhere on the page. That is close to worthless for semantic claims, because the words survive an inversion of the meaning. Inverting "a duplicate term in your file **overrides** the bundled entry" to say the opposite left the string `**overrides**` present and the suite green. The same held for the pkuseg name-versus-path rule, and for a wrong environment-variable name in a table that passed because the correct name appeared elsewhere on the page.

Three changes fixed it:

1. **Each rule is stated once, as a table**, which the test parses and runs row by row against the implementation. pkuseg rows go through `_local_pkuseg_model` against a module whose `available_models` is built from the row's own column. Dictionary-precedence rows are compared against behaviour measured per backend, with jieba observed through a stub tokenizer recording `load_userdict` and `add_word`.
2. **The unbound prose restating each rule was deleted.** A second, unexecuted copy can always be inverted on its own.
3. **The tables must discriminate.** A table collapsed to one uniform outcome fails, so the guard cannot go vacuous the way its predecessor did.

### Mutation battery

Every documented rule was mutated and the killing test recorded:

| Mutation | Killed by |
| --- | --- |
| Invert the pkuseg name-versus-path resolution table | `test_documented_pkuseg_resolution_rule_is_executed` |
| Invert the dictionary-precedence table | `test_documented_dictionary_precedence_rule_is_executed` |
| Wrong env var in the provisioning table | `test_documented_hanlp_env_var_and_pkuseg_asymmetry_are_real` |
| Claim construction does not touch the filesystem | `test_documented_failure_timing_table_is_executed` |
| Reframe the jieba skip reason as a missing extra | `test_documented_skip_reasons_are_quoted_verbatim` |
| Mislabel `boundary_f1` as liveness / hardware-dependent | `test_documented_metric_gating_table_matches_the_shipped_suite` |
| Claim a `chars_per_second` throughput gate | `test_documented_metric_gating_table_matches_the_shipped_suite` |
| State either two-layer threshold as the other | `test_documented_two_threshold_split_matches_the_suite` |
| Falsely claim the blob stub is rejected by the floor | `test_documented_floor_scope_limit_is_pinned_by_measurement` |
| Change a dictionary limit (`100000` to `99999`) | `test_documented_dictionary_limits_match_the_defaults` |
| Read the guide with `decode("cp1252")` instead of UTF-8 | `test_guide_reads_as_utf8_on_every_platform` |
| Truncate the guide to zero bytes | all tests fail |

The guard also caught two real defects in the guide's own first draft: a `SpanToken` import from a module that does not re-export it, and a claim that dictionary merge is "first occurrence wins" for every backend, which is false for jieba.

## Windows portability fix

The first CI run failed two of these tests on the Windows job only. Both built a doc-facing path needle with `str(...relative_to(ROOT))`, which renders `tests\unit\processing\test_zh_segmentation.py` on Windows and cannot match a Markdown document that correctly spells paths with forward slashes. Fixed with `.as_posix()` on the derived side, so the assertions still require the guide to name the exact shipped paths — made portable, not loosened.

Generally: **documentation paths are always forward-slash, so never derive a doc-facing path string from the local filesystem separator.**

### There was no encoding bug

The same CI output showed a replacement character where the guide has an em dash, which reads like a `cp1252` decode. It is not, and the distinction is worth recording for anyone reading the same log:

```
utf-8 bytes of an em dash: e28094
decoded as cp1252 -> 'â€"'   (three characters)
decoded as utf-8  -> '—'
```

A `cp1252` misread produces **three** mojibake characters, not one. CI showed a **single** `�`, which is a correctly decoded U+2014 being re-encoded for display by the Windows console or the log pipeline. Checked before changing anything: all 14 text reads in the module already pass `encoding="utf-8"`, and the only bare `.open()` calls are `"rb"` for `tomllib`. Both CI failures are explained entirely by the path separators.

The guard was added anyway, because "no encoding bug today" is worth pinning. `test_guide_reads_as_utf8_on_every_platform` asserts the em dash and Han text survive the read, that no U+FFFD is present, and that the `â€"` mojibake signature is absent. It is mutation-tested: swapping `_guide()` to `decode("cp1252")` fails it.

One trap worth naming, since it cost a cycle: the first version of the POSIX guard inspected its own source for the forbidden pattern and matched its own assertion literals, counting three occurrences where it expected two. It now assembles the needles at runtime. A guard that can match itself will catch someone else out.

Unrelated to this branch: `tests/unit/core/test_budget.py::test_wall_time_clock_raises_after_deadline` appears on the Windows job across sibling PRs. It is pre-existing upstream code, present at `2965ec10` and touched by no commit in this stack — a `time.monotonic()` granularity issue on Windows.

## Three details worth reviewing closely

**1. The metric table separates liveness from quality.** A binary "gated" column would be wrong in one direction or the other, because the suite *does* assert all three metrics — but not in the same sense:

| Metric | Suite asserts | Role | Determinism |
| --- | --- | --- | --- |
| `boundary_f1` | `>= SEGMENTATION_BOUNDARY_F1_FLOOR` | quality gate | deterministic |
| `dictionary_hit_rate` | `== 1.0` | quality gate | deterministic |
| `chars_per_second` | `> 0.0` | liveness only | hardware-dependent |

`chars_per_second > 0.0` proves the timer ran; it judges nothing about the number. Calling that "gated" would let a reader conclude throughput is held to a standard when no throughput value can fail. The test extracts each assertion from the suite source and compares it to the cell, so the column cannot drift.

**2. One metric, two thresholds.** `boundary_f1` is asserted twice at different values because it answers two different questions. The in-repo reference segmenter is held to exactly `1.0` — a fixed implementation over a checksum-pinned corpus, so anything less means the harness or corpus broke, not that a backend is weak. Installed backends get `>= 0.90`, because their output legitimately varies by build and by user-supplied model. Only the second is a bar a user's backend must clear, and the guide says so.

The guide also documents the floor's known blind spot rather than hiding it, with both numbers measured: a character-per-token segmenter scores 0.8068 and is rejected, while one that keeps dictionary terms but merges every other run into a single blob scores 0.9722 and **passes**. Raising the floor past that point would exceed the shipped default backend's own gate. The scope is stated in #1556's own words: the suite validates the span protocol and dictionary-term survival, plus a 0.90 floor that rejects character-level output; it does not validate general segmentation quality.

**3. The Unicode-folding caveat is self-invalidating.** The exact-`1.0` threshold draws its strength from determinism, which makes it specifically fragile to Unicode folding: if a step folding width or applying NFKC were ever introduced upstream of segmentation, full-width `ＣＴ` would fold to `CT`, offsets would shift, and a genuine offsets defect could score as a pass.

**This is a constraint to preserve, not a present-day defect.** No such step exists: the shared input gateway strips whitespace and enforces size and encoding limits, applies no Unicode normalization, and sits on request-entry surfaces that segmentation never routes through.

The caveat is not prose. `test_documented_unicode_folding_caveat_still_holds` asserts that `openmed/utils/gateway.py` stays free of `unicodedata` and all four normalization forms, that the full-width probe round-trips through `normalize_text` unchanged, and that `zh_segmentation` imports neither `normalize_text` nor `utils.gateway`. It is mutation-tested in both directions: adding `import unicodedata` to the gateway fails it, and flipping the guide to claim NFKC *is* applied also fails it. If anyone wires folding upstream of segmentation, this documentation gets rewritten rather than quietly becoming false.

That battery was **re-run after the portability fix**, not assumed to still hold: both directions still fail as intended, and both mutated files were restored and diff-verified. The guard reads `gateway.py` and `zh_segmentation.py` with explicit `encoding="utf-8"`, so this evidence does not rest on a platform-dependent read.

## Verification

Measured at `2965ec10` against the canonical baseline.

| | Baseline `2965ec10` | This branch |
| --- | --- | --- |
| Failed | 31 | 31 |
| Passed | 9816 | 9873 |
| Skipped | 86 | 90 |

**Sorted FAILED sets are identical (31/31).** The 31 are pre-existing and environmental: 21 `test_zh_normalize` and 4 `test_zh_segmentation` from OpenCC and jieba not being installed locally, 3 `test_brand_system`, and one each in `test_api_surface_diff`, `test_mcp_entrypoint`, `test_model_integrity`.

The `+57` passed reconciles exactly: `9816 + 29` (this PR's documentation tests) `+ 28` (#1556's conformance tests that pass in this environment) `= 9873`. The `+4` skipped are #1556's live-model cases, which skip with an explicit reason naming the missing asset.

- **Documentation and nav gates:** `test_zh_segmentation_docs`, `test_docs_publication`, `test_docs_language_coherence`, `test_troubleshooting_docs`, `test_zh_segmentation_conformance` — 81 passed, 4 skipped (29 + 11 + 3 + 10 + 28).
- **pre-commit:** all hooks pass on the changed files.
- **Strict documentation build:** not runnable in this environment — `mkdocs-static-i18n` is absent. Confirmed pre-existing and content-independent by reproducing the identical `Config value 'plugins': The "i18n" plugin is not installed` error against unmodified upstream with these changes stashed. The nav and publication-parity gates that the strict build would exercise do pass.


maziyarpanahi:

PR #2037, opened 2026-08-02
feat(eval): verify India surrogate consistency and register the clinical de-identification suite

### LINKED ISSUE #1580: Verify India surrogate consistency and register the clinical de-identification suite
## Parent
Parent: #1503

## Summary
Complete OM-677 by verifying multi-script surrogate stability, registering the India suite in the standard harness, and documenting its safety boundary.

## Scope
- Verify that the transliterated same-person identity in the corpus maps to exactly one stable surrogate across documents and scripts.
- Register the India suite alongside the existing policy-compliance and SHIELD suites in the standard eval harness.
- Surface combined per-label policy coverage, leakage, and surrogate-consistency verdicts.
- Document synthetic-only provenance, assist-only non-decisional use, local/offline execution, and the exclusion of real hospital data.

## Acceptance criteria
- The cross-document fixture produces exactly one surrogate identity for all declared aliases.
- The standard harness discovers and runs the India suite without special caller wiring.
- One report exposes policy coverage, zero-leak, and consistency verdicts without raw PHI.
- Documentation states the synthetic-only and assist-only safety boundary.
- `.venv/bin/python -m pytest tests/ -q` is green.

## Dependencies
- All earlier OM-677 child issues.
- Coordinate with #1498 for stable India-aware replacement behavior.

## Out of scope
- Real-data benchmarking.
- Model training or leaderboard work.
- Non-clinical Indian PII domains.

---
Parent: #1503
Task: OM-677-child-4
Size: M

PouyanJay:
## Description

Completes OM-677 by verifying multi-script surrogate stability for the synthetic India clinical de-identification corpus, registering the India suites in the standard eval harness, and documenting the safety boundary.

Closes #1580

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which does not break existing functionality)
- [ ] Breaking change
- [x] Documentation update
- [x] Test addition/improvement

## Changes Made

- Added `openmed/eval/suites/india_surrogate_consistency.py`: a cross-document, cross-script surrogate-consistency gate plus a combined report.
- Registered `india_clinical_phi_leakage` and `india_surrogate_consistency` in `DEFAULT_SUITES`, with loader and metadata branches.
- Added `run_india_clinical_suite_report`, one report exposing DPDP per-label policy coverage, the residual zero-leak verdict, and the surrogate-consistency verdict.
- Documented synthetic-only provenance, assist-only non-decisional use, and local/offline execution in both the corpus README and the India onboarding pages.

## What "consistency" means here

Every declared alias of one synthetic person resolves to a single surrogate identity across documents and across Latin, Devanagari, and Tamil; the vault stores exactly one surrogate and returns the identical surface when asked again; each rendered surrogate stays in its source script; no alias surface survives, judged by canonical fold and edit similarity rather than exact substring; an identifier repeated across documents keeps one surrogate that still satisfies its own shape or checksum validator; and generated synthetic negative identifiers do not collide. Every run executes twice with a fresh vault and is compared to reject non-determinism.

## Two checks are built so they cannot pass by construction

**1. `surrogate_count` reported coverage that did not exist.** An earlier revision computed the "one surrogate per identity" claim as `len({...})` over values fetched from the vault store. A store miss returns `None`, and `len({None}) == 1`, so a vault that returned **zero** surrogates for **three different people** still reported `surrogate_count: 1, passed: True`. The metric was also not independent of the identity check: once the aliases share a key it can only mirror `identity_count`. Both causes are fixed. `None` is never collected; store misses are counted separately as `missing_surrogate_count`; and each alias is resolved a second time and must return the byte-identical surface. Under a store whose writes silently no-op, the gate now reports `missing=3, surrogate_count=0, passed=False` while `identity_count` correctly still reads `1`.

**2. Identifier surrogates are drawn from a per-run counter, not derived from the source value.** Seeding the generator from `hmac(source)` made every surrogate a pure function of its input, so the values matched regardless of whether the vault had linked anything. Drawing from a run counter makes cross-document reuse depend on the vault actually linking the occurrences.

## Known divergences: recorded, never gated

Both are reported in every result with `gated=False` and guarded by a test, so a future fix breaks loudly rather than letting a stale record rot.

**Divergence 1 — transliteration-aware matching does not fold across scripts.** The opt-in `transliteration_aware_name_matching` path does not collapse the corpus aliases into one identity. `canonical_indic_name_key` returns `indic-name-v1:aarav sharma` (Latin), `indic-name-v1:aarava sharma` (Devanagari inherent vowel), and `indic-name-v1:aarav charma` (Tamil "ச" rendered "ch"), so `indic_names_match` is `False` for all three pairings of the same synthetic person. The gated verdict covers the default matching path, which is the shipped default.

**Divergence 2 — the same Aadhaar in a Hindi note and a Tamil note draws two different surrogates.** Personal names are keyed with the language folded away (`key_lang='india'`), but structured identifiers keep the document language in the vault key:

```
key_for('<synthetic Aadhaar>', label='ID_NUM', lang='hi') vs lang='ta'
  same text_hash: True      keys equal: False
key_for('<synthetic name>', label='PERSON', lang='hi') vs lang='ta'
  keys equal: True          key lang: 'india'
```

Names link across languages; structured identifiers do not. This was invisible under the original source-derived seeding described above, which returned matching values while the vault had never linked them. The gated linkage check is therefore scoped to `(value, language)` — the granularity the vault actually guarantees, so ABHA, which repeats within a single language in this corpus, is genuinely gated and fails closed. The cross-language gap is recorded rather than enforced because choosing the right normalization for structured identifiers is a design decision beyond this issue's scope.

## Alias-leak threshold: measured, not guessed

Exact-substring containment is too weak, since a surrogate that is a near-transliteration of the real name is trivially re-identifiable while sharing no exact substring. Fold-equality alone does not catch it either, because the repository's own canonical fold deliberately keeps `arav sarma` and `aarav sharma` distinct. Detection therefore falls back to edit similarity over canonical folds, calibrated on the shipped corpus:

| | similarity |
|---|---|
| near-miss `Arav Sarma` vs alias | 0.960 |
| near-miss Devanagari vs alias | 0.941 |
| real surrogate vs Latin alias | 0.655 |
| real surrogate vs Devanagari alias | 0.643 |
| real surrogate vs Tamil alias | 0.778 |

Threshold `0.85` sits between the two clusters with roughly 0.07 margin below and 0.09 above.

## Post-rebase re-verification

Re-measured numerically after rebasing onto `2965ec10`, since that commit introduced a shared input-normalization gateway:

- The three canonical folds are unchanged, and all three pairwise `indic_names_match` results are still `False`.
- Identifier keys are still language-scoped and `PERSON` keys still normalize to `key lang='india'`; the asymmetry is neither fixed nor worsened.
- Near-misses still 0.960 / 0.941 against real surrogates at 0.655 / 0.643 / 0.778, so the 0.85 threshold keeps its margin on both sides.
- The gate still reports `passed=True` with both divergences recorded.

The gateway does not interact: `normalize_text()` is a size and encoding guardrail (strip plus character and byte limits) and performs no Unicode or linguistic normalization, and it never reaches `canonical_indic_name_key`. `surrogate_vault.py`, `indic_name_match.py`, and `anonymizer/` contain no gateway references; `core/pii.py` imports only `validate_language`. Verified empirically: `normalize_text` returns all three aliases byte-identical.

## Scope limits

- Negative probes are **structured identifiers only**. This suite generates no negative name surfaces and therefore does not gate name collisions; those are covered by the existing `indic-name-consistency` suite.
- Only the first of the two determinism runs contributes verdicts. A divergence visible solely in the second run registers as non-determinism, but its per-verdict detail is not recorded.

## Deviations from the issue's file list

Every file the issue did not list, with the gate forcing it:

| File | Gating test |
|---|---|
| `tests/unit/eval/test_metrics.py` | Exact-tuple assertion on `DEFAULT_SUITES` at line 39; registering the two suites requires updating it. |
| `docs/onboarding-india.hi.md` | `tests/unit/test_docs_publication.py` enforces nav and localized-nav parity. The EN and HI pages are structurally mirrored, so updating one without the other leaves the localized page inconsistent. |

Considered and confirmed **not** required:

- `scripts/release/public_api_baseline.json` — additions are non-breaking; `tests/unit/release/test_public_api_compat.py` only fails on unannounced breaks, and it is green.
- `openmed/eval/golden/loader.py` `_SPECIALIZED_FIXTURE_NAMES` — already excludes the India corpus from generic golden discovery.

## Public API changes

Renamed within symbols introduced by this branch: `NEGATIVE_PERSONA_COUNT` to `NEGATIVE_IDENTIFIER_COUNT`, `NEGATIVE_PERSONA_COLLISION` to `NEGATIVE_IDENTIFIER_COLLISION`, and serialized `negative_persona_count` to `negative_identifier_count`. The former names described name-collision coverage that the suite does not provide. Everything else is additive; nothing pre-existing was removed or changed.

## Data provenance

All data is synthetic and algorithmically generated. **No real PHI appears anywhere in this change.**

This matters especially here because the suite exercises Indian national identifiers. Every Aadhaar, ABHA, PAN, and phone value used or produced is generated at runtime by the shared India provider registry (`generate_aadhaar`, `generate_abha_number`, `generate_pan`, `generate_indian_phone`) and checked against the corresponding shape or checksum validator. **No real Aadhaar, ABHA, or PAN number is present in the corpus, the tests, or the generated surrogates.** Values that satisfy a Verhoeff or structural checksum are valid in *form* only; they are not issued identifiers and correspond to no person.

The corpus contains no production data, no DUA-restricted content, and no real hospital data. It is an assist-only, non-decisional evaluation fixture, not clinical ground truth, and must not be used to make patient-care decisions. Execution is local, offline, and deterministic.

## Verification

Rebased onto `2965ec10`; `git log 2965ec10..HEAD` shows exactly one commit.

| | failed | passed | skipped |
|---|---|---|---|
| canonical baseline at `2965ec10` | 31 | 9816 | 86 |
| this branch | 31 | 9839 | 86 |

Sorted `FAILED` sets are **identical** — zero new, zero missing. The passed delta is exactly the 23 tests added by this PR (9816 + 23 = 9839). The 31 pre-existing failures are unrelated to this change (`test_zh_normalize`, `test_zh_segmentation`, `test_brand_system`, `test_api_surface_diff`, `test_mcp_entrypoint`, `test_model_integrity`).

Pre-commit: all hooks pass on all eight changed files, including ruff check, ruff format, bandit, and the secret scan.

Harness registration was verified by execution rather than inspection: both suites load through the corpus loader as standard `BenchmarkFixture` objects, and `run_benchmark` scores `exact_span_f1 = 1.0` against a gold-echo runner for each.

## Suggested follow-ups

Not filed, for the maintainer to open if they agree:

1. **India transliteration-aware name matching does not fold the Devanagari inherent vowel or Tamil "ச" across scripts.** Matching is strict canonical-key equality in `openmed/core/indic_name_match.py` with no similarity fallback, so the folds must agree exactly. Deliberately not fixed here: it touches shared core matching used well beyond this corpus, and lowering `DEFAULT_INDIC_NAME_SIMILARITY_THRESHOLD` to force the fold raises collision risk for genuinely distinct names, which is the failure mode the conservative fold exists to prevent.
2. **Structured identifier surrogates are not linked across languages in code-mixed records.** Consider normalizing the key language for non-name labels, as `_source_identity` already does for personal-name labels. This affects any multi-language record set, not only India.

## Checklist

- [x] I have added tests that prove the feature works
- [x] New and existing unit tests pass locally
- [x] I have updated the documentation accordingly
- [x] I have added docstrings to new functions and classes
- [x] I ran the formatting and lint gates


maziyarpanahi:

PR #2038, opened 2026-08-02
Expand CBLUE evaluation coverage and Chinese model operations guidance

### LINKED ISSUE #1561: Expand CBLUE evaluation coverage and Chinese model operations guidance

PouyanJay:
## Relationship to #1560 — read first

**This branch contains #1560's commit `8ec072bb` ("feat: add opt-in Chinese clinical NER local-asset conformance"). #1560 must merge first.** This PR shows two commits and closes one issue; that is deliberate stacking, not scope creep.

Both changes touch `openmed/eval/suites/chinese_clinical_ner.py` and `tests/unit/eval/test_chinese_clinical_ner.py`. Sending them to `master` independently would defer the identical conflict to merge time, so this branch was rebased onto #1560 instead.

**Rebase promise:** once #1560 merges, I rebase onto the new `master` and this PR narrows to my commit alone.

### Conflict note for whoever lands second

Both conflicts were **additive-only** export/import collisions — #1560 added three CamelCase classes, this PR added the `MULTILINGUAL_FALLBACK_MODEL` constant. Resolved by keeping both sides and preserving the file's existing ordering convention (ALL_CAPS constants sorted, then CamelCase classes sorted). No logic conflict.

**Whoever lands second should re-run `tests/unit/eval/test_chinese_clinical_ner.py` in full rather than trusting the merge.** Two sets of fixture helpers can merge cleanly in text and still be semantically incompatible. That file is where #1560's test growth lives. I ran it standalone after resolving and before continuing the rebase: **35 passed, 1 skipped**. I also verified beyond the conflict markers that both sides' function bodies survived and that there are **zero duplicate `def`/`class` names**, so no helper is shadowed. The 6 `SpanValidationWarning`s in that run are #1560's intentional negative-path conformance cases, not a defect.

---

## Summary

Extends the user-supplied benchmark interface to the CBLUE task shapes that carry clinical entity annotations, without implementing CMeIE relation decoding.

- **Interface extension.** Adds a `row_adapter` hook to `load_multilingual_ner_benchmark` so a task whose native release layout is not text-plus-spans can normalize its rows while reusing the shared license boundary, offset validation, and canonical label mapping.
- **Two task shapes.** CHIP-CDN (diagnosis-term normalization) and IMCS-V2-NER (medical-dialogue entity recognition), each with its own path variable and access-controlled license entry. Both native release shapes are accepted: `sentence` as a string or character list, `BIO_label` as a whitespace-joined string or tag list, CHIP-CDN standard terms split on `##`.
- **New suite** `cblue-task-coverage`: per-task exact span metrics, per-label counts, CHIP-CDN normalization accuracy, and a fail-closed provenance gate.
- **CMeIE stays out of scope** and is *enforced*, not merely intended: `cblue_task_shape("cmeie")` and `load_cblue_task("cmeie")` both raise, covered by `test_cmeie_is_not_registered`.
- **Operations guidance**: tested local-asset layouts, evaluation commands, report interpretation, upgrade/rollback, and model-card evidence requirements.

## The defect this PR's own review surfaced

The provenance report **misreported in both directions**, which is worse than having no report because a provenance report is trusted:

- `OPENMED_CBLUE_PATH` was **silently ignored** while the report claimed `"configured": true`. A user who provisioned a licensed corpus, set the documented variable, and ran the documented command got a green provenance report over six invented rows, and their licensed data was never read — the exact failure this suite advertises preventing.
- The mirror image: running over real licensed data via explicit `paths={...}` reported `"status": "skipped"`, `"reason": "OPENMED_CHIP_CDN_PATH is not set"` while reporting F1 over that corpus.

**Fix.** Path resolution now honours explicit argument → task variable → per-task child of `OPENMED_CBLUE_PATH` → synthetic fallback, mirroring the existing CMeEE precedent. Each fixture is tagged with its `source_kind` at load, and the report's availability block is rewritten from the fixtures that actually reached the scorer — so a report describes the *run*, not the loader's configuration. Both directions are now **verified rather than asserted**: with the variable set, `user data read: True` and availability `configured`; on a synthetic run, `configured: false, status: "synthetic"` with the reason naming the variable that was not set.

Two further correctness fixes from the same review:

- **Unknown task crashed instead of failing closed.** The `unknown_task` reason code was documented but unreachable — the scoring loop raised from the task-shape lookup first, under both `fail_on_provenance=True` and `False`, defeating the documented escape hatch for obtaining a report instead of an exception. Unknown tasks are now skipped in scoring and recorded as a finding.
- **Fixture id collision.** CBLUE numbers rows per file, so two task files both starting at `"1"` collided with an unactionable `ValueError`. Ids are namespaced `"{task}/{id}"` with the original preserved as `source_record_id`.

Worth recording as process: **the second defect masked the first.** Reproducing the availability bug hit the id collision first. Two defects stacked, and the outer one would have hidden the inner one from anyone who fixed only what a review named — an argument for reproducing each finding before fixing it.

### The same failure mode, a second time (commit `3473cd78`)

Windows CI caught the no-vendor guard building its scanned-path set with `str(Path.relative_to(...))` and comparing it against forward-slash literals:

```
str(rel)       -> 'openmed\\eval\\fixtures\\cblue_chip_cdn_synthetic.jsonl'   # != literal, fails
rel.as_posix() -> 'openmed/eval/fixtures/cblue_chip_cdn_synthetic.jsonl'      # fixed
```

This is not merely a one-line test fix, and it is worth stating why. The guard failed on Windows **whether or not the invariant held** — so on that platform it proved nothing while looking like coverage. That is precisely the failure mode this PR exists to fix in the provenance report: **a report that describes something other than what it actually verified.** The report claimed `"configured": true` without reading the configured data; the guard claimed to have checked committed payloads when it had only failed a string comparison. Two instances of one lesson in a single change is worth naming rather than hiding.

Fixed by normalising the *actual* side with `as_posix()` rather than loosening the comparison, which would have traded a false failure for a false pass.

**Both halves of the guard were re-proved after the change**, because a path fix could plausibly have broken the non-vacuity built earlier: injecting a row with `synthetic: false` still fails it, and renaming a fixture out of the regex pattern still fails it. The fixture was restored byte-identical afterwards. Re-proving is the difference between a fix and a hope.

## Data provenance

**All synthetic, algorithmically generated. No real PHI, no real clinical Chinese text, and no CBLUE record is read, copied, or paraphrased.**

CBLUE is access-controlled (Tianchi dataset 95414); CHIP-CDN's standard-term side additionally derives from a separately licensed national clinical ICD-10 edition. Nothing is bundled, downloaded, cached, or redistributed. Repository-internal paths are rejected so licensed data cannot be committed by accident.

The two committed fixtures are composed from a closed **invented** vocabulary of enumerator morphemes (jia/yi/bing, used the way English uses A/B/C) joined to generic shape words. Character offsets and per-character BIO tags are computed from the composed segments, not hand-written.

The generator is committed at `scripts/benchmarks/generate_cblue_synthetic_fixtures.py` so the provenance claim is **reproducible rather than asserted**. It is deterministic *by construction* — it draws no random numbers and reads no clock, environment, or filesystem state, so there is no seed to pin — and `--check` verifies the committed fixtures byte for byte:

```
$ python scripts/benchmarks/generate_cblue_synthetic_fixtures.py --check
committed CBLUE synthetic fixtures match the generator
```

`test_committed_fixtures_are_reproducible_from_the_generator` enforces this continuously.

### Scope limit of the no-vendor guard, stated plainly

`test_no_real_benchmark_payload_filenames_are_committed` checks the `metadata.synthetic` **self-attestation, not file content**. A file containing real rows marked `synthetic: true` would pass it. The generator test is what actually content-locks the two committed fixtures. This limitation is documented in the guard's own docstring.

The guard was also made **non-vacuous by construction**: it now asserts that every expected benchmark-named payload was actually in scope, so renaming a fixture out of the regex pattern fails the build instead of silently skipping it. Both halves were verified by deliberate mutation — a row with `synthetic: false` fails it, and renaming a fixture out of scope fails it.

The issue lists no Files section, so every file below is derived. Each extra is named with the gate that forces it.

| File | Why it is required |
| --- | --- |
| `tests/unit/eval/test_metrics.py` | Exact-tuple gate on `DEFAULT_SUITES` at line 39 — hard-fails on any new suite |
| `tests/unit/eval/test_license_aware_ner_suites.py` | No-vendor guard regex widened to the new task names, plus the scope assertion making it non-vacuous |
| `openmed/eval/datasets/licenses.py` | `test_license_registry_covers_cblue_cmeee_and_naamapadam` requires license entries per dataset key |
| `openmed/eval/datasets/__init__.py`, `openmed/eval/suites/__init__.py`, `openmed/eval/__init__.py` | Export blocks; the registry dispatch and `DEFAULT_SUITES` entry live here |
| `openmed/eval/datasets/multilingual_ner.py` | The `row_adapter` extension point and the two new source registrations |
| `openmed/eval/suites/chinese_clinical_ner.py` + its test | Stale model-notice correction (below) |
| `docs/eval-harness.md` | **New top-level `## CBLUE task coverage` section.** Flagged explicitly so the shape is visible at a glance; it stays clear of #1560's "Optional local model assets" subsection |
| `docs/onboarding-china.md`, `docs/onboarding-china.zh.md` | Both are navigated pages with a cross-link gate in `tests/unit/test_docs_publication.py`; editing one without the other breaks it |
| `scripts/benchmarks/generate_cblue_synthetic_fixtures.py` | Not required by any gate — added so the synthetic-data claim is reproducible |

**Deliberately not added:** no new `docs/*.md` file. `tests/unit/test_docs_publication.py` requires every `docs/*.md` to appear in exactly one of `mkdocs.yml` nav or `docs/brand/system/publication.yml` classification; extending existing pages avoids that double registration.

### Stale model evidence corrected

`chinese_clinical_ner_metadata()` asserted that `zh` was "a documented multilingual routing placeholder, not a dedicated Chinese clinical NER checkpoint", and a test pinned that string. That was no longer true: `openmed/core/language_pack_catalog.py` routes `zh` to `OpenMed/OpenMed-PII-Chinese-BigMed-Large-560M-v1`, and `DEFAULT_MODEL_PLACEHOLDER_LANGUAGES = frozenset({"ru"})` shows `zh` is not a placeholder.

Replaced the fixed claim with a `model_evidence` block **derived** from the catalog, so it tracks the code in both directions rather than restating a fact that can go stale. Documented model-card requirements for both the multilingual-fallback case and a substituted dedicated local model, including the requirement to disclose any overlap between a model's training corpus and the CBLUE split it is scored on.

## Verification

Three commits: `8ec072bb` (#1560), `3525033d` (this feature), `3473cd78` (the POSIX path fix).

Canonical baseline `SHARED_BASELINE_2965ec10_FAILED.txt` at `2965ec10`: **31 failed / 9816 passed**.

This branch: **31 failed / 9876 passed / 87 skipped / 1 xfailed**.

**The FAILED sets are identical** — set difference is empty in both directions, so this branch introduces no regression and fixes none. The totals are **unchanged by `3473cd78` as expected**: a path-normalisation fix adds no tests, so only the platform on which the guard is meaningful changes, not the counts.

The `+60 passed / +1 skipped` delta reconciles exactly, and includes #1560 because this branch contains it. Collected tests under `tests/unit/eval`: 1163 at `2965ec10` → 1186 at `8ec072bb` (+23 from #1560) → 1224 at this commit (+38 from this PR). The +23 lands as +22 passed and +1 skipped, because #1560's local-asset test skips without local assets.

Other checks:

- `tests/unit/eval/test_chinese_clinical_ner.py` standalone after conflict resolution: **35 passed, 1 skipped** (6 `SpanValidationWarning`s are #1560's intentional negative-path cases)
- `tests/unit/eval`: 1219 passed, 6 skipped
- Docs gates (`test_docs_publication.py`, `test_pages_manifest.py`, `test_docs_language_coherence.py`): **47 passed**
- **pre-commit on all changed files: every hook passed** — trailing whitespace, end-of-file, large files, case conflicts, merge conflicts, debug statements, mixed line ending, ruff check, ruff format, bandit, secret detection
- Generator `--check`: byte-identical
- Test-gate coverage: all 11 documented gate reason codes have tests, plus a docs-drift test binding the documented list to the implementation so the two cannot diverge
- Metrics are not tautological: an imperfect runner yields recall `7/13 ≈ 0.538` at precision 1.0, a mislabelling runner shows scoring is label-sensitive, and a wrong normalizer yields accuracy `2/3`

### Cross-platform path audit (confirmed negative)

After the Windows failure, the rest of the module and the CBLUE suites were audited for the same shape — any filesystem path compared to a literal, or embedded in a message a test matches on. No other instance:

- Every `pytest.raises(match=...)` string checked for separators: `"relation decoding"`, `"explicit local"`, `"inside the repository"`, `"does not exist"`, `"duplicate fixture id"` — none contains one
- Remaining `str(path)` uses classified: `str(path).strip()` in `cblue.py` is a truthiness check on a possibly-empty value, and the generator's `drifted` entry is display-only with no test matching on it — both separator-agnostic
- Fixture ids `"{task}/{id}"` confirmed to be **constructed strings with a literal `/`**, not filesystem paths, so `"chip-cdn/1"` is stable cross-platform
- `benchmark_name.search(path.name)` operates on a bare filename

### Windows CI: the other failure is not from this PR

A Windows CI run shows a second failure, `tests/unit/core/test_budget.py::test_wall_time_clock_raises_after_deadline`. **It is pre-existing upstream code, not ours, and was deliberately not touched.** It is present at `2965ec10` and absent from this PR's diff. The cause is clock granularity: it sets `max_wall_time=0.001` and sleeps 10 ms, but Windows `time.monotonic()` has ~15.6 ms granularity, so the deadline can read as not yet elapsed. It is confirmed absent from this branch's POSIX failure set. Fixing an unrelated upstream test inside this PR would be scope creep, so it is reported separately.

### Not verified locally

`make docs-build` **could not be run** in this environment: the docs extras are unavailable offline (`uv` cannot fetch `babel==2.17.0` / `mypy==2.3.0` from cache). The three docs gate suites above were run instead (47 passed). **This acceptance criterion is left explicitly unverified rather than claimed.**

### Packaging

Measured packaged-tree delta via `git archive <rev> openmed | wc -c`: 11,079,680 → 11,120,640 bytes = **+40,960 bytes**, against 264,520 bytes of wheel headroom in `gates/release_budgets.json` (~15%). An earlier figure of 2,780 bytes in my own working notes was wrong — it counted only the fixtures and omitted 32,415 bytes of new Python; corrected here.

Minor: the `DEFAULT_SUITES` entry is inserted mid-tuple rather than appended. Cosmetic — nothing indexes that tuple.

Closes #1561


maziyarpanahi:

PR #2015, opened 2026-08-02
feat: register Indic and Urdu user-supplied PII languages

PouyanJay:
Closes #1569

Order 2 of 5 in epic #1488 (OM-662). Depends on #1568, which is closed (merged as #1573).

## Summary

Registers the Indic and Urdu routing candidates (`mr`, `ne`, `bn`, `as`, `ta`, `kn`, `ml`, `gu`, `pa`, `or`, `ur`) across the public language surfaces, while keeping model selection explicit for the codes that ship no bundled weights.

Mapping to the issue's scope items:

- **Register the eleven codes in the supported-language, display-name and model-prefix surfaces.** Nine were already registered by #1573; `ne` and `ur` gain display names, model prefixes, and REST / MCP / TypeScript / Go language enums.
- **Distinguish user-supplied-only languages from languages with bundled default PII models, and return an actionable error when no model is supplied.** `USER_SUPPLIED_MODEL_LANGUAGES` becomes a first-class acceptance set, joined into every union that already joins `INDIC_NER_LANGUAGES` / `NATIONAL_ID_ONLY_LANGUAGES`.
- **Native or documented approximate Faker locale resolution, without inventing national-ID providers.** `ne` resolves to Faker's native `ne_NP` (verified present in Faker 40.15.0), so it is neither a documented approximation nor a conceptual locale needing a backend. No national-ID provider is invented for it. `ur` keeps its existing `ur_PK` locale and CNIC provider.
- **Update generated language enums and source-of-truth tests.** `docs/api/openapi.json`, `openmed/interop/tools.json` and `tests/fixtures/language_pack_adapters.json` are regenerated from their generators, never hand-edited.

## Please read first: most of this issue was already satisfied

Nine of the eleven codes needed no work. `USER_SUPPLIED_MODEL_LANGUAGES = {gu, kn, ml, ne, pa, ur}` already landed with #1573, and `LANGUAGE_NAMES` / `LANGUAGE_MODEL_PREFIX` / `LANG_TO_LOCALE` already carried Marathi, Bengali, Assamese, Tamil, Odia, Gujarati, Kannada, Malayalam and Punjabi.

| code | before this PR | after |
|---|---|---|
| `mr` `bn` `as` `ta` `or` | fully registered, default model resolves | unchanged |
| `gu` `kn` `ml` `pa` | registered; actionable error without a model; accepted with one | unchanged |
| `ur` | no display name or model prefix; accepted by `pii.py` but **rejected** by `pipeline.py` | registered end-to-end |
[...]

PR #1852, opened 2026-07-21
feat: add Urdu RTL PII language pack with Pakistani CNIC validation

AlyanPremani05:
# Pull Request

## Description
Adds Urdu (`ur`) internationalization (i18n) support for PII detection, de-identification, and surrogate generation. This includes rule-based safety sweep patterns for Pakistani CNICs (National IDs), phone numbers, dates, street addresses, and postcodes, along with locale registration and test coverage.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition/improvement

## Changes Made
- Added `_URDU_PII_PATTERNS` in `openmed/core/pii_i18n.py` including Pakistani CNIC detection (`national_id`) supporting both Western and Eastern Arabic-Indic digits, as well as dashed (`XXXXX-XXXXXXX-X`) and undashed forms.
- Added Urdu contextual cues for `date`, `phone_number`, `street_address`, and `postcode` (tuned address pattern boundaries to stop at sentence delimiters like `۔` without swallowing postal codes).
- Registered `ur` in `LANGUAGE_PII_PATTERNS`, `NATIONAL_ID_ONLY_LANGUAGES`, and mapped `ur` -> `ur_PK` locale in `openmed/core/anonymizer/locales.py`.
- Implemented CNIC Faker surrogate provider in `openmed/core/anonymizer/providers/clinical_ids.py` producing format-correct 13-digit Pakistani CNICs.
- Added synthetic Urdu golden fixture in `openmed/eval/golden/ur.jsonl` and unit tests in `tests/unit/test_pii_i18n.py`.

## Testing
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have tested this change with different models/inputs

## Documentation
- [ ] I have updated the documentation accordingly
- [x] I have added docstrings to new functions/classes
- [ ] I have updated the CHANGELOG.md

## Code Quality
- [ ] I ran `make format`, `make lint`, and `make format-check`
- [ ] For Swift/OpenMedKit changes, I ran `make format-swift` and `make lint-swift`
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings

[...]

Issue #1570, opened 2026-07-13
Add grapheme-aligned mixed-script run routing


maziyarpanahi:
Parent epic: #1488
Depends on: #1568, #1569

## Summary
Evolve mixed-script segmentation into a grapheme-safe routing layer while preserving exact source offsets.

## Scope
- Segment text only at extended grapheme-cluster boundaries.
- Emit a routing decision per run covering script, candidate or selected language, normalizer, tokenizer, and numeral set.
- Preserve exact source slices and provide a compatibility path for existing tuple consumers.
- Keep neutral characters attached deterministically without losing text.

## Acceptance criteria
- Concatenating every [start, end) slice reconstructs the source exactly.
- Every start and end is an extended grapheme boundary.
- Routing metadata is deterministic and complete for Latin and all nine Brahmi scripts.
- Existing RTL rendering and normalization callers remain green.
- The full test suite passes.

Size: L
Order: 3 of 5

Issue #1569, opened 2026-07-13
Register Indic and Urdu user-supplied PII languages


maziyarpanahi:
Parent epic: #1488
Depends on: #1568

## Summary
Extend the public language registry for the Indic and Urdu routing candidates while keeping model selection explicit for languages that have no bundled default.

## Scope
- Register mr, ne, bn, as, ta, kn, ml, gu, pa, or, and ur in the supported language, display-name, and model-prefix surfaces.
- Distinguish user-supplied-only languages from languages with bundled default PII models and return an actionable error when no model is supplied.
- Add native or documented approximate Faker locale resolution without inventing national-ID providers.
- Update generated language enums and source-of-truth tests required by the public API.

## Acceptance criteria
- Every new code is accepted when a caller supplies a model and no default model is claimed for it.
- Existing hi and te defaults are unchanged.
- Locale coherence and client parity gates pass.
- The full test suite passes.

Size: L
Order: 2 of 5

PR #2039, opened 2026-08-02
feat: disambiguate Urdu from Arabic in script runs

### LINKED ISSUE #1571: Disambiguate Urdu from Arabic in script runs
Parent epic: #1488
Depends on: #1570

## Summary
Route Urdu text in the shared Arabic Unicode blocks without weakening Arabic routing.

## Scope
- Define a reviewed set of Urdu-specific code points and lexical signals.
- Add a deterministic per-run Arabic-versus-Urdu heuristic with an Arabic fallback.
- Surface the selected language in run routing metadata.
- Cover presentation forms and combining marks without logging source text.

## Acceptance criteria
- The Urdu fixture routes to ur and the Arabic regression fixture routes to ar.
- Ambiguous runs use a documented deterministic fallback.
- Offsets and grapheme boundaries are unchanged by disambiguation.
- The full test suite passes.

Size: M
Order: 4 of 5

PouyanJay:
Closes #1571

Order 4 of 5 in epic #1488 (OM-662). Routes Urdu text inside the shared Arabic Unicode blocks without weakening Arabic routing.

## Relationship to other PRs — read first

This branch is **stacked**, so it shows **3 commits while closing one issue**:

| commit | PR / issue | owner |
|---|---|---|
| `573a09a8` | #2015 — register Indic and Urdu user-supplied PII languages (#1569) | mine, open |
| `2bfb5775` | #1570 — grapheme-aligned mixed-script run routing | mine, open |
| `7e08f5f1` | **this PR** (#1571) | mine |

**Merge order: #2015, then #1570, then this.** Only `7e08f5f1` is under review here; the first two are already under review in their own PRs.

**Rebase promise:** if either parent changes during review, or `master` moves, I will reparent this branch onto the updated tip and re-verify against the shared baseline before asking for another look. Each reparent replays only my own commit (`git rebase --onto <new-parent-tip> <previous-parent-tip>`), verified by asserting the commit count above the new base.

Related but **not** in scope: #1520 (Urdu PII pack) is the maintainer's own branch. See the AC deviation below, which exists specifically to stay out of its way.

## Summary

- Defines a reviewed set of Urdu code points: the six letters not used by Arabic or Persian — tteh `ٹ` U+0679, ddal `ڈ` U+0688, rreh `ڑ` U+0691, noon ghunna `ں` U+06BA, heh doachashmee `ھ` U+06BE, yeh barree `ے` U+06D2 — plus their Arabic presentation forms.
- Adds `urdu_language_evidence()`, a deterministic per-run score, mirroring the existing `assamese_language_evidence()` precedent for the Bengali/Assamese pair.
- Evidence reorders the run's candidate list so `ur` leads; runs report `stdlib:urdu-cues` when an Urdu pack is registered, and fall back to Arabic under `stdlib:arabic-fallback` at a lower confidence when none is.
- Presentation forms are **derived at import time** from single-character NFKC decompositions, not hand-listed.
- Offsets and grapheme boundaries are untouched: `segment_by_script` and `_script_for_char` are not modified.

## The rule, and what it does on each input

| input | evidence | `candidates` | `language` | `source` | confidence |
|---|---|---|---|---|---|
| pure Arabic | 0 | `('ar','ha','ur')` | `ar` | `stdlib:script` | 0.99 |
| pure Urdu | ≥1 | `('ur','ar','ha')` | `ur` if an Urdu pack is registered, else `ar` | `stdlib:urdu-cues` / `stdlib:arabic-fallback` | 0.99 / 0.8 |
| Persian | **0** | `('ar','ha','ur')` | `ar` | `stdlib:script` | 0.99 |
| Arabic + one Urdu letter | 1 | `('ur','ar','ha')` | as pure Urdu | as pure Urdu | as pure Urdu |
| extended digits, no letters | 0 | unchanged | unchanged | unchanged | unchanged |

**Persian safety.** Persian shares peh/tcheh/jeh/gaf `پ چ ژ گ` with Urdu but uses **none** of the six letters, and it uses the same extended Arabic-Indic digits U+06F0–U+06F9. So digits **reinforce an existing letter signal but never trigger one**. Without that asymmetry, `پچگ` + `۱۲۳۴` would score as Urdu. This is deliberately more conservative than counting digits whenever an Arabic letter is present, and it matches #1520's own framing, which routes on letters and uses digits only for validator normalization.

**Confidence semantics.** 0.99 when the evidence-preferred language has a pack; **0.8 with `stdlib:arabic-fallback`** when it does not. The lower value keeps the fallback visible in run metadata rather than reading as a confident Arabic match, and it propagates into the length-weighted `DocumentLanguageDecision.confidence`.

## The `ha` (Hausa) determination

`ha` sits in the Arabic hint tuple but is national-ID-only: the input gateway rejects it when `include_national_id=False`, which is what every public edge uses. Since this PR selects within that candidate list, promoting `ha` into `language` would hand back a code the gateway refuses — the bug #2015 just fixed, reintroduced from the other direction.

**I tried to make `ha` win rather than reasoning it away, and it can be selected — so the flat claim "unreachable" is false.** The real guarantee is *ordering*:

| pack set | Arabic text | Urdu text | Persian |
|---|---|---|---|
| `(ar, ha)` | `ar` | `ar` | `ar` |
| `(ar, ha, ur)` | `ar` | `ur` | `ar` |
| `(ha, ur)` | **`ha`** | `ur` | **`ha`** |
| `(ha,)` | **`ha`** | **`ha`** | **`ha`** |

`ha` wins **only when `ar` is absent from the pack set**, which cannot happen through any shipped path, for two independent reasons:

1. `ha` has **no `LanguagePack` at all**, and `_select_pack`'s candidate tuple is built from *registered packs*, not from hint codes — so `ha` never enters it.
2. `ar` is a **built-in pack**, therefore always present; and the evidence branch moves **only** `ur`, preserving `ar` ahead of `ha` in `('ur','ar','ha')`, so the rank loop returns at `ar` at the latest.

Reaching `ha` requires a caller-injected `LanguageRouter(packs=…)` that deliberately omits `ar` — a test or embedding construction, never the gateway path. Pinned by `test_urdu_disambiguation_never_resolves_the_hausa_candidate`, which registers `ar`+`ha`+`ur` together and asserts `run.language != "ha"` and `candidates.index("ar") < candidates.index("ha")` across Arabic, Urdu, Persian and Urdu-with-extended-digits.

**Honest caveat:** `run.candidates` **does** advertise `ha` — `('ar','ha','ur')`. That is #1570's advisory field fed by the existing hint tuple, unchanged here, and the test asserts it stays. A consumer treating `candidates` as gateway-submittable would hit the 422. **`language` is the safe field.**

**1. "The Urdu fixture routes to `ur`" is proven with an injected pack, not by registering one.**

Satisfying it literally requires a `ur` `LanguagePack`, which forces deleting `ur` from `NATIONAL_ID_ONLY_CAPABILITIES` and `USER_SUPPLIED_MODEL_LANGUAGES` — contradicting merged #1852, and colliding head-on with the maintainer's reserved #1520 branch, which does exactly those two edits. `test_routing_only_languages_do_not_claim_bundled_models` asserts `USER_SUPPLIED_MODEL_LANGUAGES` is disjoint from bundled models, so the collision is enforced by a test, not just convention.

Instead the evidence and candidate reordering ship so that `_select_pack` yields `ur` **the instant #1520 registers a pack, with no further change to the router**, and the fixture gate proves it with an injected pack — mirroring `test_assamese_cues_disambiguate_the_shared_bengali_script`, which injects a `bn` pack the same way. Until then the documented Arabic fallback carries the run.

**I am happy to defer this issue entirely to #1520 if you would rather it land as one change** — the scope split here is a proposal, not an assumption.

**2. The discriminator lives in `script_detect`, not in `LanguagePack.routing_markers`.**

`_routing_marker_count` does whole-token matching, which cannot express letter-level evidence, and markers live on a pack this PR is deliberately not creating. `script_detect` is where the Bengali/Assamese analog already puts it.

**3. Files beyond the two obvious source files, each with the gate forcing it.**

| file | why |
|---|---|
| `openmed/core/script_detect.py` | the evidence function and candidate reordering |
| `openmed/core/language_router.py` | per-script source table and the fallback tier |
| `tests/unit/core/test_script_detect.py` | evidence unit tests and adversarial cases |
| `tests/unit/core/test_language_router.py` | routing, confidence tiering, the `ha` guard |
| `tests/unit/eval/test_golden_fixtures.py` | **forced by the analog's own gate** at `test_assamese_cues_disambiguate_the_shared_bengali_script` — the Bengali/Assamese pair is proven fixture-driven there, so the Arabic/Urdu pair has no equivalent proof without a mirrored test |
| `CHANGELOG.md` | repo convention for feature PRs |

**Deliberately not touched**, each verified rather than assumed: `language_pack_catalog.py` and `tests/fixtures/language_pack_adapters.json` (the Arabic hints are already `('ar','ha','ur')` and `ur` is already routable, so `script_language_hints` is byte-identical and `test_indic_and_arabic_script_language_hints_are_exact` passes unchanged); `docs/languages.md` (`ur` is already a documented row); `openmed/eval/golden/fixtures/i18n/ur.jsonl` (adding digit rows is #1520's own acceptance criterion and would pull in the leakage and bidi gates); `USER_SUPPLIED_MODEL_LANGUAGES` and `NATIONAL_ID_ONLY_CAPABILITIES` (#1520's surface).

**4. A pre-existing mislabelling fixed in passing.** The candidate-reorder branch returned a hard-coded `"stdlib:assamese-cues"` for *any* script whose hints were reordered. Once Arabic reorders, Urdu-cued runs would have been labelled with the Assamese source. The label is now per-script; Bengali keeps `stdlib:assamese-cues`, pinned by the existing gate and re-asserted by `test_bengali_evidence_keeps_its_established_routing_source`.

## Data provenance

**All synthetic, algorithmically generated, never real clinical text or PHI.** Every Arabic-script sample in the new tests is constructed in code from code points (`chr(0x0679)`, `"".join(chr(c) for c in …)`), never pasted. The presentation-form set is derived by enumerating FB50–FDFF and FE70–FEFF through `unicodedata.normalize`, not transcribed. The two golden fixtures used (`ur.jsonl`, `ar.jsonl`) are pre-existing and **unmodified**; they are read, not rewritten. No fixture offsets were computed by hand — offset equality is asserted through `segment_by_script` and `snap_span_to_grapheme_boundaries`.

## Verification

Canonical shared baseline `SHARED_BASELINE_2965ec10_FAILED.txt` at `2965ec10`: **31 failed, 9816 passed**, 86 skipped, 1 xfailed.

This branch at `7e08f5f1`: **31 failed, 9937 passed**, 86 skipped, 1 xfailed.

Failure-set diff — `comm -13` (new) **empty**, `comm -23` (fixed) **empty**, `diff` **identical, 31 == 31**. No new failures; the 31 are pre-existing and unrelated (Chinese normalization/segmentation, brand assets requiring Pillow, API-surface diff, MCP entrypoint, model integrity).

Passed-count reconciliation, `9937 − 9816 = 121`: **#2015 (33) + #1570 (73) + this PR (15)**.

`pre-commit run --files …` over all six changed files: **all hooks pass** (ruff check, ruff format, bandit, secret detection, whitespace, EOF, large files, case conflicts, merge conflicts, debug statements, line endings).

**15 new tests**, all passing:

| acceptance criterion | test |
|---|---|
| Urdu fixture → `ur`, Arabic → `ar` | `test_urdu_cues_disambiguate_the_shared_arabic_script` |
| documented deterministic fallback | `test_urdu_fixtures_fall_back_to_arabic_until_an_urdu_pack_ships`, `test_urdu_evidence_without_a_registered_pack_falls_back_to_arabic` |
| offsets and graphemes unchanged | `test_urdu_disambiguation_leaves_script_runs_and_graphemes_unchanged`, `test_urdu_disambiguation_preserves_fixture_offsets_and_graphemes` |
| presentation forms | `test_urdu_presentation_form_evidence_is_pinned_to_sixteen_forms` |
| Persian false positive | `test_persian_text_yields_no_urdu_evidence` |
| Koranic ligature trap | `test_koranic_stop_sign_ligatures_are_not_urdu_evidence` |
| digit-only discrimination | `test_extended_arabic_indic_digits_reinforce_but_never_trigger_urdu` |
| candidate reordering | `test_urdu_evidence_moves_ur_ahead_of_ar_without_dropping_candidates`, `test_urdu_cues_select_ur_over_ar_once_an_urdu_pack_is_registered` |
| confidence tiering arithmetic | `test_arabic_fallback_confidence_weights_the_document_decision` |
| gateway-invalid code never selected | `test_urdu_disambiguation_never_resolves_the_hausa_candidate` |
| Assamese source not regressed | `test_bengali_evidence_keeps_its_established_routing_source` |
| letter set is the only letter evidence | `test_urdu_exclusive_letters_are_the_only_letter_evidence` |

### Adversarial cases worth calling out

**The Koranic ligature trap.** `U+FDF0` (SALLA) and `U+FDF1` (QALA) NFKC-decompose *through* yeh barree `ے` U+06D2, but occur in **Arabic** religious text. A naive NFKC expansion of presentation forms would misroute it to Urdu. Restricting derivation to **single-character** decompositions excludes both, and the test asserts U+06D2 *is* present in their NFKC — so the filter is demonstrably what excludes them, not an accident of the range.

**The presentation-form count is pinned at exactly 16** so a Python Unicode-database upgrade fails loudly instead of silently widening or narrowing the evidence set.

**Behavioural re-verification after the final rebase.** 16 probes measured against the rebased tree rather than read from source: U+FDF0/U+FDF1 confirmed to contain yeh barree under NFKC *and* score 0; exactly 16 presentation forms; Persian 0 with and without extended digits; digits alone 0, Arabic stem + digits 0, `ے` + three digits → 4; `0.8`/`stdlib:arabic-fallback` versus `0.99`/`stdlib:script`; Assamese still `stdlib:assamese-cues` with `('as','bn')`.

## Known scope, stated rather than left to be found

- **"Urdu-exclusive" in the identifier names and docstrings overclaims.** `ٹ ڈ ڑ ے` are shared with Punjabi Shahmukhi, Saraiki and Kashmiri; `ھ` with Sindhi, Uyghur and Kashmiri. The accurate claim is **"not used by Arabic or Persian"**, which is the only property the rule needs, since `ar` and `ur` are the sole Arabic-script candidates with packs. Text in those languages will score as Urdu — acceptable while no packs exist, and worth revisiting if one is added.
- **`ۓ` U+06D3 and `ہ` U+06C1 are as diagnostic as the six and currently score zero.** Deliberate: the set is frozen at the six that both this issue and #1520 name, so the discriminator matches the specification under review rather than expanding it unilaterally. A cheap recall win for #1520 to fold in with its pack.
- **Once a `ur` pack is registered, plain Arabic drops from 0.99 `stdlib:script` to 0.8 `stdlib:pack-priority`** — not caused by this change, but by a second Arabic pack ending the single-candidate short-circuit. Whoever lands #1520 should expect it.


maziyarpanahi:

PR #1573, opened 2026-07-13
Add Indic Unicode script routing metadata

maziyarpanahi:
# Pull Request

## Description
Adds the Unicode script registry and candidate-language foundation for Indic and Urdu routing. This is the first independently mergeable child of epic #1488 and deliberately does not bundle or claim default models for the new routing-only languages.

## Type of Change
- [x] New feature (non-breaking change which adds functionality)
- [x] Test addition/improvement

## Changes Made
- Add Bengali, Gurmukhi, Gujarati, Odia, Tamil, Kannada, and Malayalam Unicode ranges alongside existing Devanagari and Telugu support.
- Expand candidate-language hints for all nine Brahmi scripts and add Urdu as an Arabic-script candidate.
- Declare new routing-only languages separately from bundled-model languages and add exact regression coverage.

## Testing
- [x] I have added tests that prove the feature works
- [x] New and existing unit tests pass locally with my changes
- [x] `.venv/bin/python -m pytest tests/ -q` — 4624 passed, 48 skipped
- [x] `make format`, `make lint`, and `make format-check`
- [x] Scoped pre-commit hooks

## Documentation
- [x] No user-facing documentation update is required for this internal foundation slice

## Code Quality
- [x] I ran `make format`, `make lint`, and `make format-check`
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings

## Dependencies
- [x] I have not added any new dependencies

## Checklist
- [x] I have read the contributing guidelines
- [x] My commits have clear, descriptive messages
- [x] I have organized the commit as one focused change

## Related Issues
Closes #1568
Related to #1488

## Screenshots/Examples
Not applicable.

Issue #1568, opened 2026-07-13
Add Indic Unicode script registry and routing language hints


maziyarpanahi:
Parent epic: #1488

## Summary
Add the Unicode script taxonomy and candidate-language metadata that later mixed-script routing builds on. This is the first independently mergeable slice of the OM-662 epic.

## Scope
- Add Bengali, Tamil, Kannada, Malayalam, Gujarati, Gurmukhi, and Odia to the supported script registry and Unicode ranges alongside Devanagari and Telugu.
- Expand candidate language hints to Devanagari (hi, mr, ne), Bengali (bn, as), Arabic (ar, ur), and the single-script Indic languages.
- Declare the new Indic and Urdu codes as user-supplied routing candidates without adding bundled default models.
- Preserve existing Latin, Hindi, and Telugu detection behavior.

## Acceptance criteria
- Each of the nine Brahmi scripts is classified by detect_script using synthetic text.
- candidate_languages_for_script returns the exact candidates for every added script and Arabic includes Urdu.
- Existing Latin, Devanagari, and Telugu regressions remain green.
- No per-language model is bundled and the full test suite passes.

## Files
- openmed/core/script_detect.py
- openmed/core/pii_i18n.py
- tests/unit/core/test_script_detect.py

Size: M
Order: 1 of 5

Issue #1488, opened 2026-07-09
Indic script auto-detection and per-script routing layer covering all nine Brahmi scripts plus Perso-Arabic (Urdu) with mixed-script run segmentation


maziyarpanahi:
## Summary
OpenMed's `script_detect.py` only carries Devanagari and Telugu ranges and maps both Devanagari and Telugu to single languages, so Bengali/Assamese, Tamil, Kannada, Malayalam, Gujarati, Gurmukhi, Odia and Perso-Arabic Urdu are invisible and mixed-script clinical notes are mis-routed. A per-script router that segments text into script runs and dispatches each to the right normalizer, tokenizer, numeral folder, and language model is the backbone the rest of this initiative plugs into. This is hard because Devanagari serves several languages, Bengali script serves Bengali and Assamese, Urdu shares the Arabic block with Arabic, and run boundaries must preserve exact grapheme-aligned offsets.

## Scope
- [ ] Extend `openmed/core/script_detect.py` with Unicode block ranges and `SUPPORTED_SCRIPTS` entries for Bengali, Tamil, Kannada, Malayalam, Gujarati, Gurmukhi, Odia and Perso-Arabic (Urdu) alongside existing Devanagari/Telugu.
- [ ] Update `SCRIPT_LANGUAGE_HINTS` so each script maps to its candidate languages (Devanagari→hi/mr/ne, Bengali→bn/as, Arabic block→ar/ur disambiguation).
- [ ] Add Urdu (`ur`) and the new Indic language codes to `openmed/core/pii_i18n.py` `SUPPORTED_LANGUAGES`/name/prefix maps as user-supplied-model entries, keeping hi/te defaults intact.
- [ ] Make `segment_by_script` emit grapheme-aligned run boundaries and a routing decision per run (normalizer, tokenizer, numeral set, language).
- [ ] Add Arabic-vs-Urdu disambiguation using Urdu-specific code points and a per-run heuristic.
- [ ] Provide a synthetic mixed-script clinical note fixture exercising all eleven scripts/languages and assert routing.

## Acceptance criteria
- [ ] `detect_script` and `segment_by_script` correctly classify and bound runs for a synthetic note mixing all nine Brahmi scripts plus Urdu and Latin.
- [ ] Offset-correctness gate: concatenating all run `[start,end)` slices reconstructs the source byte-for-byte and every run boundary is a grapheme boundary.
[...]

PR #2040, opened 2026-08-02
feat: add grapheme-aligned mixed-script run routing

### LINKED ISSUE #1570: Add grapheme-aligned mixed-script run routing

PouyanJay:
Closes #1570

Order 3 of 5 in epic #1488 (OM-662).

## Relationship to other PRs — read first

This branch contains **two commits** while closing **one issue**:

| commit | belongs to |
| --- | --- |
| `573a09a8` | PR #2015 (#1569, Indic and Urdu user-supplied PII languages) |
| `2bfb5775` | this PR (#1570) |

#1570 declares `Depends on: #1568, #1569`. #1568 is merged (PR #1573). #1569 is **PR #2015, still open**, so its commit is carried here as the base.

- **PR #2015 must merge first.** Once it does, only `2bfb5775` remains and this PR becomes a single-commit change. I will rebase and force-update this branch after #2015 lands, and again on request if `master` moves.
- **#1571 (Urdu vs Arabic disambiguation, order 4) stacks on top of this branch** and depends on the public surface below.

Both commits are rebased onto `master` at `2965ec10` (PR #2023, the shared input-normalization gateway). `git rev-list --count HEAD..master` is 0.

## Summary

Evolves mixed-script segmentation into a grapheme-safe routing layer while preserving exact source offsets, per the issue's scope:

- **Segment only at extended grapheme-cluster boundaries.** A run can no longer split a combining sequence, an Indic virama conjunct, a zero-width joiner sequence, or a regional-indicator pair.
- **Emit a routing decision per run.** `LanguageRun` gains `candidates`, `normalizer`, `tokenizer`, and `numeral_set`.
- **Preserve exact source slices with a compatibility path.** `segment_by_script` now yields `ScriptRun`, a `NamedTuple` that is a real `tuple` subclass, so all eight existing call sites keep unpacking and comparing against plain tuples unchanged.
- **Keep neutral characters attached deterministically.** Each cluster takes the script of its first script-bearing code point, so a cross-script combining mark stays with the base character it decorates.

### The defect, demonstrated

Segmenting by code point placed boundaries inside grapheme clusters whenever a combining mark carried a different script from its base:

| input | before | grapheme boundaries | after |
| --- | --- | --- | --- |
| `a` + U+0951 + `b` | `(0,1,Latin) (1,2,Devanagari) (2,3,Latin)` | `{0,2,3}` | `(0,3,Latin)` |
| `Ra` + U+093C + `i` | `(0,2,Latin) (2,3,Devanagari) (3,4,Latin)` | `{0,1,3,4}` | `(0,4,Latin)` |
| U+0995 + U+0951 | `(0,1,Bengali) (1,2,Devanagari)` | `{0,2}` | `(0,2,Bengali)` |

Devanagari and Tamil virama conjuncts, emoji ZWJ sequences, regional-indicator pairs and ZWNJ sequences were already handled correctly and are retained as regression guards at zero violations.

Boundary snapping was considered and rejected: snapping `Ra` + U+093C + `i` assigns the cluster `a` + U+093C to Devanagari even though its base is Latin. Taking the script of the cluster's first script-bearing code point is the rule that keeps the mark with its base.

## Security: quadratic segmentation removed

The first implementation of the grapheme pass resolved cluster starts with a backward walk. Inside a cluster whose extending marks carry a different script from the base, **every mark re-entered the walk and rewound to the cluster start**, making segmentation O(n·k) — quadratic in cluster length.

This was reachable through untrusted input. `MAX_PII_COMBINING_SEQUENCE` and `MAX_PII_FORMAT_SEQUENCE` reset on each other's characters, so a payload of the shape `"a" + (U+0951 * 63 + ZWJ) * N + "b"` passes `validate_pii_input()` and then reaches `segment_by_script`, `route_runs`, `normalize_for_pii_detection`, `is_indic_text`, `is_han_dominant`, `refine_privacy_filter_span`, `rtl_render`, and `eval/metrics` — several of them on the same document per request.

The fix memoizes the resolved cluster start: offsets already known to share a cluster stop the walk on contact, so rewinds telescope and total boundary probes stay within `n + (script changes)`. Segmentation is linear.

Measured on this branch after the rebase, which confirms by measurement rather than inspection that the fix survived the replay:

| payload | before | after | growth per doubling |
| --- | --- | --- | --- |
| 1,922 B | 0.440 s | 0.0024 s | — |
| 3,842 B | 1.739 s | 0.0048 s | 2.00x |
| 7,682 B | 7.038 s | 0.0094 s | 1.94x |
| 15,362 B | — | 0.0194 s | 2.06x |

On the 7,682-byte payload, `route_runs` goes from 14.3 s to 0.0097 s and `is_indic_text` from 11.8 s to 0.0096 s.

`test_segment_by_script_stays_linear_on_adversarial_combining_runs` pins this. It counts boundary probes rather than wall-clock time, so it is deterministic under CI load — a timing threshold would flake and then be weakened until it meant nothing. It is non-vacuous: the pre-fix implementation issues **3,225,600 probes** against the gate's limit of 10,248 (315x over); the shipped implementation issues **2,559**.

### Benign-path cost

Interleaved best-of-N against the pre-change implementation on an idle machine, so the comparison is like-for-like:

| document (4 KB) | pre-change | this PR | ratio |
| --- | --- | --- | --- |
| Latin only | 0.771 ms | 0.796 ms | 1.03x |
| mixed 4-script | 3.271 ms | 3.665 ms | 1.12x |
| Latin + Devanagari | 1.950 ms | 2.135 ms | 1.09x |
| Indic heavy | 7.284 ms | 7.809 ms | 1.07x |

Grapheme correctness costs 3–12% on ordinary text. Cluster boundaries are resolved only at offsets where the script changes, so single-script text performs no boundary probes at all.

## Correctness evidence

A randomized differential compared the new segmentation against the previous one across 60,000 generated strings mixing all supported scripts with combining marks, joiners, Hangul jamo, regional indicators, emoji, ideographic description sequences and neutrals:

- **17,660 of 60,000 strings produced different output.**
- **In 17,660 of 17,660 of those, the previous output violated a grapheme-cluster boundary and the new output violated none.**

The change therefore alters behaviour only where the previous behaviour was already wrong. A second differential of 120,000 strings against an independent cluster-iteration reference, plus 4,000 longer strings of 40–160 characters that exercise the memo far harder, produced **0 mismatches** and **0** reconstruction, boundary or tiling failures.

## Public surface

Frozen contract that #1571 builds on:

```python
class ScriptRun(NamedTuple):          # tuple subclass; existing unpackers unaffected
    start: int; end: int; script: str

@dataclass(frozen=True, slots=True)
class LanguageRun:
    start, end, script, language, confidence, source   # unchanged
    candidates: tuple[str, ...] = ()                   # added
    normalizer: str = "unicode-defense"
    tokenizer: str = ""
    numeral_set: str = "ascii"
```

Also added: `SCRIPT_NORMALIZERS`, `SCRIPT_NUMERAL_SETS`, `normalizer_for_script`, `numeral_set_for_script`. The normalizer table is derived from the Brahmi script tuple in `openmed.processing.text` under an alias, so it cannot drift from the normalizer that actually runs; the alias exists because `script_detect` exports its own `INDIC_SCRIPTS`, and an unaliased import would silently rebind that public name. `test_indic_script_sets_agree_across_modules` pins the two sets so the alias cannot be collapsed.

### Caveat for consumers of `candidates`

`candidates` is **advisory**. It lists the ordered routing hints for the run's script; `language` may fall outside it when a neighbouring script, a routing marker, or the optional language-ID backend selects a different pack, and `source` records which path decided.

Concretely: `candidate_languages_for_script("Arabic")` returns `('ar', 'ha', 'ur')`, but `ha` belongs to `NATIONAL_ID_ONLY_LANGUAGES` without being in `USER_SUPPLIED_MODEL_LANGUAGES`, so the shared gateway rejects it at `include_national_id=False` — the setting every public edge uses. **`language` is the gateway-submittable field; `candidates` is not.** #1571 has pinned a test that its Urdu/Arabic disambiguation never resolves to `ha`.

The issue lists no files, so every path below is stated with the gate that forces it.

| file | why |
| --- | --- |
| `openmed/core/script_detect.py` | `ScriptRun`, grapheme-aligned `segment_by_script`, the two routing tables and their accessors. |
| `openmed/core/language_router.py` | The four `LanguageRun` fields and their population. |
| `openmed/core/__init__.py` | Re-exports `segment_by_script` and `LanguageRun`; `ScriptRun` and the new accessors must join them. |
| `openmed/core/decoding/spans.py` | **Not implied by the issue.** Adds `grapheme_break_checker`, the constant-time boundary predicate the linear fix requires; reaching into the module's private helpers instead would have been worse. Forced by `test_segment_by_script_stays_linear_on_adversarial_combining_runs`. The same file's `_find_structured_match` annotation moves from `list` to `Sequence` because `list` is invariant, so `list[ScriptRun]` does not satisfy `list[tuple[int, int, str]]` — surfaced by mypy, fixed rather than suppressed. `grapheme_break_checker` is deliberately **not** re-exported from `openmed.core.decoding`, keeping the package's public surface unchanged. |
| `tests/unit/core/test_script_detect.py` | Exact-list expectation at the mixed-script segmentation test, and `test_script_language_hints_cover_detectable_scripts`, which forces any new script-keyed mapping to cover `set(SUPPORTED_SCRIPTS) \| {UNKNOWN_SCRIPT}`. |
| `tests/unit/core/test_language_router.py` | `test_runs_tile_mixed_script_input_without_gaps_or_overlaps` and the cross-script routing accuracy gate. |
| `tests/unit/core/test_grapheme_span_decoding.py` | Consistency gate proving `grapheme_break_checker` agrees with `iter_grapheme_cluster_spans` on every offset, including IDS, Hangul jamo, regional-indicator and CRLF cases. |
| `tests/unit/processing/test_indic_normalizer.py` | One expectation changed; see below. |
| `CHANGELOG.md` | Repository convention for new exports and user-visible behaviour changes. |

### The one changed expectation

`test_normalization_offset_map_round_trips_multiscript_phi_spans` expected `(16, 19)` for the Gurmukhi token and now expects `(15, 19)`.

Offset 16 is U+0A71 GURMUKHI ADDAK, category `Mn`, preceded by a SPACE at offset 15. Under UAX #29 rule GB9 the mark joins the preceding character, so **offset 16 is not a grapheme-cluster boundary** and the previous code was splitting the cluster — precisely the defect this issue exists to fix. The only grapheme-legal boundaries are 15 and 17; 17 would route a Gurmukhi mark into the preceding Bengali run. 15 is therefore the correct value, and only 1 of the 9 tokens in that fixture is affected.

The expectation is pinned as a **literal**. Deriving it from the grapheme helpers would share the break rules with the code under test, so a regression in those rules would move both sides together and still pass. Generating the literal table also surfaced a transcription error in the fixture's Odia token, which ends U+0B5F, not U+0B2F.

## Data provenance

All test data is synthetic and algorithmically generated in code from explicit Unicode code points. No real clinical text, no real or realistic PHI, and no third-party corpora. The randomized differentials build their inputs by sampling code points at a fixed seed. Every asserted offset was computed by script and verified for grapheme-boundary membership.

## Verification

Full suite on this branch, compared against the canonical baseline captured at `master` = `2965ec10`:

| --- | --- | --- | --- | --- |
| baseline `2965ec10` | 31 | 9816 | 86 | 1 |
| this branch `2bfb5775` | 31 | 9922 | 86 | 1 |

**Sorted FAILED sets are 31/31 identical — zero new failures, zero incidentally fixed.** The 31 pre-existing failures are environmental (missing `opencc` and `jieba` extras, a Sigstore bundle, brand SVG byte-portability, an MCP console entry point) and unrelated to this change. The +106 passing tests are the 18 added here plus those arriving with the rebased base.

`pre-commit run --files` over every changed file: all hooks pass — trailing whitespace, end-of-file, large files, case conflicts, merge conflicts, debug statements, mixed line endings, `ruff check`, `ruff format`, `bandit`, and secret detection. None of the repository's known pre-existing hook failures are in this change's file set. `mypy` on the touched modules reports only the one pre-existing error that is present on the base commit.

## Observation for maintainers

There is currently **no performance gate anywhere in `tests/`**. This PR adds the first one, covering `segment_by_script` only. Every other hot path remains unguarded, so a change reintroducing this class of complexity cliff elsewhere would land with CI green. The probe-counting approach used here is cheap to replicate if that is worth generalizing.


maziyarpanahi:

PR #2069, opened 2026-08-03
fix: measure request wall-time budgets with perf_counter

PouyanJay:
## Summary

`BudgetClock` measures elapsed wall time with `time.monotonic()`. On Windows that
clock is backed by `GetTickCount64()`, whose resolution is **15.6 ms**, until
CPython 3.13 switched it to `QueryPerformanceCounter()` ([gh-88494](https://github.com/python/cpython/issues/88494)).
`requires-python = ">=3.10"`, and the CI matrix runs Windows on 3.11 and 3.12, so
every currently supported Windows configuration is affected.

This changes the two clock reads in `openmed/core/budget.py` to
`time.perf_counter()`.

## Production impact

This is a correctness bug in the budget feature itself, not only a flaky test.
On Windows + Python 3.10-3.12:

1. **Any `max_wall_time` below ~15.6 ms is silently unenforceable.** `elapsed`
   cannot observe an overrun shorter than one tick, so a request budget that
   should trip never does. It fails open — the request runs unbounded.
2. **Longer budgets trip early.** `elapsed` jumps from 0 to 15.6 ms the moment a
   request straddles a tick boundary, so a budget can be reported as exhausted
   after microseconds of real work.

Neither mode raises anything visible; the budget just silently does the wrong
thing.

## Evidence

Simulating the Windows clock pair (`monotonic()` quantized to 15.6 ms,
`perf_counter()` exact) and sweeping the phase within the tick at which the
request starts, over 1000 start phases per row:

**(A) False negative — work that overran the budget by 2x and must always trip:**

| `max_wall_time` | real elapsed | budget missed (before) | after |
|---|---|---|---|
| 1 ms | 2 ms | **87.2 %** of start phases | 0 % |
| 2 ms | 4 ms | **74.4 %** | 0 % |
| 5 ms | 10 ms | **36.0 %** | 0 % |
| 10 ms | 20 ms | 0 % | 0 % |

**(B) False positive — work well inside the budget that must not trip:**

| `max_wall_time` | real elapsed | wrongly tripped (before) | after |
|---|---|---|---|
| 10 ms | 0.5 ms | **3.2 %** of start phases | 0 % |
| 10 ms | 1 ms | **6.4 %** | 0 % |
| 10 ms | 4 ms | **25.6 %** | 0 % |

## Why `perf_counter()` is the right clock

- It is documented as "a clock with the highest available resolution to measure
  a short duration" — exactly this use case. On Windows it is
  `QueryPerformanceCounter()` (sub-microsecond) on every supported version.
- It is monotonic on all supported platforms
  (`time.get_clock_info("perf_counter").monotonic is True`).
- `monotonic()` was not chosen for any semantics this code depends on. Both
  clocks have an undefined reference point where only differences are valid,
  and a difference is all `BudgetClock.elapsed` computes.
- No cross-process comparability is needed: `started_at` is produced by
  `RequestBudget.start()` and consumed by `BudgetClock.elapsed` in the same
  process, and is never serialized, persisted, or compared across processes.
  `BudgetClock` is constructed in exactly one place. (`perf_counter` is in fact
  system-wide, so this is not a regression even in principle.)

## Regression tests and mutation proof

Two tests in `tests/unit/core/test_budget.py` substitute a `_WindowsTickClock`
that reproduces the Windows pair on any host — `monotonic()` quantized to a
15.6 ms tick, `perf_counter()` exact — so the property is pinned on Linux and
macOS too rather than being left to be discovered on Windows CI:

- `test_wall_time_budget_is_enforced_below_the_windows_tick` — a 2 ms budget must
  still trip after 8 ms of work that fits inside a single tick.
- `test_wall_time_budget_does_not_trip_early_on_a_tick_boundary` — crossing a
  tick boundary after 1 ms of work must not be read as 15.6 ms and must not trip
  a 10 ms budget.

Mutation-tested. Reverting both call sites to `time.monotonic()` and rerunning
the file:

```
FAILED tests/unit/core/test_budget.py::test_wall_time_budget_is_enforced_below_the_windows_tick
  - Failed: DID NOT RAISE BudgetExceededError
FAILED tests/unit/core/test_budget.py::test_wall_time_budget_does_not_trip_early_on_a_tick_boundary
  - BudgetExceededError: wall-time budget of 0.01s exceeded (elapsed 0.015625s)
2 failed, 33 passed
```

The remaining 33 tests in the file pass under the mutation, so the two new tests
are precisely targeted at the clock and nothing else.

## Changes to existing tests

No assertion was loosened. Two existing fixtures
(`test_wall_time_clock_raises_after_deadline` and
`test_batch_processor_wall_time_budget_is_fresh_per_item`) install a fake `time`
module that previously exposed only `monotonic()`. They now expose both
`monotonic()` and `perf_counter()` returning the same fake value, so each test
keeps pinning its own property (deadline raising; per-item clock freshness)
without being coupled to which clock the implementation reads. The local
variable was renamed `monotonic_time` -> `fake_now` to match.

## Verification

- Full suite before (clean worktree at `de90aba7`): **31 failed, 9971 passed**, 88 skipped, 1 xfailed.
- Full suite after: **31 failed, 9973 passed**, 88 skipped, 1 xfailed.
- The two failure sets are **byte-identical** (`diff` of the sorted `FAILED` lines is empty). The 31 are pre-existing and unrelated (missing `fontTools` / `Pillow` optional dev deps, API-surface diff, etc.). The +2 passes are exactly the two new tests.
- `pre-commit run --files openmed/core/budget.py tests/unit/core/test_budget.py`: all hooks pass.
- `mypy`: `Success: no issues found in 4 source files`.
- `scripts/release/check_public_api.py`: exit 0, output identical to baseline. No public API change.

## Scope

Only the wall-time resolution bug. No refactors, no behavior changes beyond the
clock, no changes to any other module.

## Relationship to currently failing PRs

10 open PRs (#2027-#2040) have been red on Windows solely because of the budget
tests in this area. Worth noting for reviewers: the *test-level* symptoms were
already patched on `master` — 0d20700f made
`test_batch_processor_wall_time_budget_is_fresh_per_item` deterministic and
c02e3daf did the same for `test_wall_time_clock_raises_after_deadline` — so
those branches mainly need a rebase onto current `master`. This PR fixes the
underlying **production** defect that both of those tests were papering over,
which the rebases do not address.

This PR is independent and touches only `openmed/core/budget.py` and
`tests/unit/core/test_budget.py`; it should not conflict with any of them, and I
will rebase if it lands after other work in this area.

## Data provenance

No data involved. The regression tests use a synthetic, algorithmically driven
fake clock; no fixtures, no recorded timings, no real or derived patient data.


maziyarpanahi:

PR #2268, opened 2026-08-06
feat: add catalog coherence gate over manifest canonical_labels

### LINKED ISSUE #2246: Add a read-only catalog coherence CI gate over the manifest
## Summary
Even with a committed manifest, derived surfaces (model_registry, pii_i18n SUPPORTED_LANGUAGES/DEFAULT_PII_MODELS, README counts, mkdocs tables, model cards, canonical_labels cross-check) can drift from models.jsonl. section 3.4 (line 378) requires a read-only coherence.yml drift gate that regenerates derived surfaces FROM the committed manifest and runs 'git diff --exit-code'. It must never re-fetch the HF API, so a new org repo does not flap the gate red (that lands via the refresh PR).

## Scope
- [ ] Add scripts/manifest/regenerate_surfaces.py that rebuilds all manifest-derived surfaces from the committed models.jsonl: model_registry OPENMED_MODELS, pii_i18n.SUPPORTED_LANGUAGES + DEFAULT_PII_MODELS, README 'N models / N languages' badges, mkdocs model+benchmark tables.
- [ ] Add coherence.yml that runs the regenerator then 'git diff --exit-code' and fails on any disagreement.
- [ ] Add a canonical_labels <-> core/labels.CANONICAL_LABELS cross-check (every manifest canonical_labels value is a member of CANONICAL_LABELS).
- [ ] The gate is strictly read-only: it must not call the HF API.

## Acceptance criteria
- [ ] Running the regenerator on a clean checkout produces no git diff (green at HEAD).
- [ ] Mutating any derived surface by hand and re-running the gate makes 'git diff --exit-code' fail.
- [ ] The coherence workflow contains no HF API call (verified by inspection); it reads only the committed manifest.
- [ ] A canonical_labels-not-in-CANONICAL_LABELS row makes the cross-check fail.

## Out of scope
- Fetching/refreshing the manifest from the HF API (OM-006).
- Auto model-card generation (OM-015).

## Files
- scripts/manifest/regenerate_surfaces.py
- .github/workflows/coherence.yml
- openmed/core/pii_i18n.py
- tests/unit/core/test_catalog_coherence.py

---
Task: OM-007  ·  Milestone: v2.3  ·  Priority: P0  ·  Size: M
Depends on: OM-006  ·  Blocks: —
Roadmap: v2.3 rolling release train: On-device reach and ecosystem packaging
Spec: PLANS/V2/EXECUTION/tasks/OM-007.md

DrVelvetFog:
Addresses #2246 (OM-007).

## Scope — most listed surfaces are already coherent by construction

Before adding a regenerator I checked each derived surface the issue names against `master`; all but one are already gated:

- **`model_registry` OPENMED_MODELS** is computed at import (`OPENMED_MODELS = build_registry(_MANIFEST_ROWS, _REGISTRY_STATE)`), so it can't drift from the manifest.
- **`pii_i18n` SUPPORTED_LANGUAGES / DEFAULT_PII_MODELS** are derived from the manifest at import and the module already raises / auto-corrects on mismatch — self-validating.
- **README counts + registry model cards** already have `manifest_diff.regenerate_registry_surfaces()` + `registry_surface_errors()`, gated by `registry-coherence.yml` (#2091).
- **mkdocs benchmark tables** are regenerated and parity-checked by `scripts/docs/stage_pages.py::render_leaderboard()` in `pages.yml`.

A second regenerator for those would duplicate that logic and create a competing source of truth. The one surface with no existing check is `canonical_labels`, so this PR adds exactly that and leaves the rest to the workflows that already own them. Happy to consolidate into a single entry point if you'd prefer.

## The check

`canonical_labels` is validated against `core.labels.CANONICAL_LABELS`. The subtlety: `normalize_label()` maps any unrecognized label to `OTHER`, and `OTHER` is itself canonical — so a bare `normalize_label(x) in CANONICAL_LABELS` is true for **every** string and gates nothing. New `labels.is_recognized_label()` is true only when a label resolves through an explicit path (CMeEE / alias map / direct canonical) and false on the `OTHER` fallthrough.

So HEAD stays green while the check still bites: the alias forms `CHEM` / `SIMPLE_CHEMICAL` (166 rows) resolve to `CHEMICAL` and pass, but a genuinely unknown label fails.

Kept the manifest as-is and made the check alias-aware (option A from the issue thread). If you'd rather `canonical_labels` store canonical `CHEMICAL` outright (option B), that's a one-shot normalization I can fold in here.

## Acceptance criteria

- **Green at HEAD** — `manifest_label_errors()` returns `[]` on the committed manifest.
- **Non-member row fails** — an unrecognized value is flagged even though it normalizes to `OTHER`.
- **No HF API call** — `coherence.yml` reads only the committed manifest.
- **Tests green** — `tests/unit/core/test_catalog_coherence.py` (6 cases).

## Files
- `openmed/core/labels.py` — `is_recognized_label()` (+ `__all__`).
- `openmed/core/catalog_coherence.py` — `manifest_label_errors()` + `python -m openmed.core.catalog_coherence` gate.
- `.github/workflows/coherence.yml` — read-only catalog coherence workflow.
- `tests/unit/core/test_catalog_coherence.py`.
- `CHANGELOG.md`.

`ruff check .` / `ruff format --check` clean; label + new tests green (295), no regression.

Closes #2246


maziyarpanahi:
Maintainer review complete for exact head `02af7967f26964118e75c4bcc3716762b24f0d13`.

- Issue gate: #2246 is open on the v2.3 milestone, the PR implements the strict canonical-label validation selected in the issue, and the issue labels are mirrored on the PR.
- Validation: all 29 current-head checks passed with 3 intentional skips, including catalog coherence, the seven-platform Python matrix, lint, security, SBOM, Nix, docs, and container gates. The combined 19-PR release rehearsal also passed 13,671 tests with 116 skips.
- Merge decision: ready. Using a merge commit to preserve the single contributor-authored implementation commit.

PR #2916, opened 2026-08-20
fix(interop): preserve HL7 v2 blank lines through parse and serialize

PouyanJay:
## Summary

The nightly `Fuzz` sweep intermittently fails with:

```
FAILED tests/fuzz/test_format_parsers_fuzz.py::test_registered_format_parsers_resist_mutated_input
Failed: hl7 crashed or returned a broken offset map
AssertionError  (rendered == canonical_source)

Falsifying example:
  b'\nMSH|^~\\&|SYNTHETIC|TEST|OPENMED|LOCAL|202401011200||ADT^A01|MSG00001|P|2.5\n...'
```

The falsifying input is the ordinary seed message with a **leading newline**. The finding is real, and it is wider than the fuzz harness.

## Root cause

`HL7Message.parse` splits on the segment separator and then discards every empty entry (hl7v2.py:160):

```python
raw_segments = [segment for segment in raw_segments if segment]
```

`serialize()` rejoins only the survivors. A trailing separator was already tracked separately via `trailing_separator`, so it round-trips; leading and interior blank lines had nowhere to be recorded and were silently lost.

## Why this is not just a test artifact

`redact_hl7v2` round-trips through parse and serialize, so redaction rewrote the message layout:

```
input  lines: 4      blank line survives redaction: False
output lines: 3
```

For a de-identification library that is a quiet correctness problem: the redacted message is no longer byte-aligned with its input, so any consumer doing offset arithmetic against the original gets shifted results. Every blank line was affected, not only the leading one the fuzzer happened to find:

| input | before | after |
|---|---|---|
| leading blank line | 90 → 89 | 90 → 90 |
| interior blank line | 90 → 89 | 90 → 90 |
| extra trailing blank line | 90 → 89 | 90 → 90 |
| blank lines in every position | lossy | exact |

## Change

Record blank-line positions alongside the existing `trailing_separator` flag, and reinsert them when serializing:

- `HL7Message.blank_line_positions: tuple[int, ...] = ()`, defaulted so existing construction is unaffected.
- `parse` captures the indexes of empty entries before filtering them out.
- `serialize` reinserts them, skipping any position beyond the current segment list so a caller who mutates `segments` gets the old behaviour rather than a crash.

Blank lines carry no HL7 content and are still **not** parsed into segments, so `segments`, `segment_names()` and the `segment_index` used for narrative provenance (hl7v2_narrative.py:313) are all unchanged. `HL7Message` is not constructed anywhere outside `parse`, and nothing inserts or removes segments, so the positional representation has no other callers to disturb.

BOM handling is deliberately untouched: `parse` still strips a leading BOM without restoring it, which is the behaviour the fuzz harness already encodes as `source.lstrip("﻿")`.

## Deviations

The obvious cheaper fixes were both rejected, and it is worth saying why:

1. **Canonicalise in the fuzz test** (strip leading newlines the way the BOM is stripped). This would have turned the job green while leaving the `redact_hl7v2` data loss in place. The fuzz test was right and the parser was wrong.
2. **Reject blank lines with `ValueError`.** A two-line change, and `ValueError` is already in `_EXPECTED_REJECTIONS` (test_format_parsers_fuzz.py:94), so the sweep would have passed for free. But real HL7 feeds do emit stray blank lines, and a de-identification tool should be lenient about what it accepts and exact about what it emits. This trades silent corruption for a hard failure.

## Verification

- The exact CI falsifying example now satisfies `rendered == canonical_source`.
- `pytest tests/unit/interop/test_hl7v2.py` → 20 passed (8 tests before, 12 added across 7 parametrized round-trip cases plus segment-identity, redaction and out-of-range-position regressions).
- `pytest tests/unit/interop tests/unit/multimodal tests/fuzz` → 712 passed, 41 skipped.
- `HYPOTHESIS_PROFILE=fuzz-nightly pytest tests/fuzz/test_format_parsers_fuzz.py -q -m fuzz` no longer reports the hl7 target.
- Full suite → 12,518 passed, 114 skipped, 0 failed.
- `pre-commit run --files openmed/interop/hl7v2.py tests/unit/interop/test_hl7v2.py` → all hooks pass.

## Data provenance

All fixtures are synthetic, built in-test from the existing `tests/fuzz/seeds/minimal.hl7` shape with placeholder identifiers (`MRN12345`, `Jane Roe`). No real messages, and no data copied from any external source.

## Honest scope note

This fixes the hl7 target, but it will **not** on its own make the nightly `Fuzz` job reliably green. Running the `fuzz-nightly` profile against a cleared Hypothesis database surfaces at least two further pre-existing crashes in other parsers, both reproducible on an unmodified `master`:

- `document:eml:.eml` → `email.errors.HeaderWriteError: folded header contains newline`, on a `From:` header carrying NUL and high bytes.
- `document:odt:.odt` → `zlib.error: Error -3 while decompressing data: invalid distance too far back`, from a corrupt zip member. `zlib.error` is not covered by `_EXPECTED_REJECTIONS`.

Both are out of scope here and left untouched. Happy to open follow-ups if useful. Their existence also explains why the job fails on some nights and not others: the search finds a different one depending on the seed.

## Related

- #2676 (`alberthammerich`, Windows format-parser fuzz worker recovery) touches `tests/fuzz/test_format_parsers_fuzz.py`. This PR stays out of that file entirely, so the two are disjoint. I will rebase if it lands first.
- A sibling PR fixes the unrelated `Model Release Gates` failure. Independent, disjoint files.


maziyarpanahi:
Pouyan, thank you for tracing this to the blank-segment handling instead of papering over serialization output. I added CRLF-specific regression coverage on your branch and brought it forward onto current master after an unrelated memory-ceiling runner flake. The exact head passes 646 interop tests with 17 expected skips, scoped hooks, the full hosted matrix, and a clean merge-tree rehearsal against current master. I am squash-merging this now.

PR #2915, opened 2026-08-20
test: cover installed pandas Spark UDF annotations

PouyanJay:
## Summary

Adds a concrete compatibility regression test proving that the Spark UDF binds the installed `pandas.Series` class at runtime.

The original contribution correctly identified the pandas 3 module-path regression in the nightly gate. Current `master` now uses a stronger injected-class identity assertion for the unit contract; this repaired branch retains the contributor insight without weakening that assertion and adds a separate installed-pandas check.

## Validation

- `pytest tests/unit/interop/test_spark_udf.py -q` — 12 passed, 1 skipped
- `pytest tests/unit/interop -q` — 635 passed, 17 skipped
- Scoped pre-commit hooks passed

No dependency pins or runtime behavior changed.

PR #2917, opened 2026-08-20
fix(ci): lowercase the GHCR image reference

PouyanJay:
## Summary

Both container workflows build their image reference as `ghcr.io/${{ github.repository }}`. GHCR requires lowercase repository names, so on any repository whose owner carries uppercase characters the publish step fails:

```
ERROR: failed to build: invalid tag "ghcr.io/<Owner>/openmed:sha-3515457baa4f": repository name must be lowercase
##[error]buildx failed with: ERROR: failed to build: invalid tag ...
```

`maziyarpanahi/openmed` is entirely lowercase, so upstream never hits it. Every fork of a capitalised owner does, on every push to `master`.

## Change

One case-folding step per workflow, placed before the first `IMAGE_NAME` use in its job:

```yaml
- name: Resolve the lowercase image name
  shell: bash
  run: |
    owner_repo=$(printf '%s' "$GITHUB_REPOSITORY" | tr '[:upper:]' '[:lower:]')
    echo "IMAGE_NAME=ghcr.io/${owner_repo}" >> "$GITHUB_ENV"
```

Writing to `GITHUB_ENV` overrides the workflow-level `env:` value for every later step in the job. That matters because `IMAGE_NAME` is consumed both from the shell (`${IMAGE_NAME}`, five sites) and from an expression context (`subject-name: ${{ env.IMAGE_NAME }}`); a step output would have required rewriting all seven call sites, this required none.

`tr` rather than the `${VAR,,}` expansion, so the step does not depend on bash 4.

## Scope: two workflows, not one

`image-signing.yml:22` carries the identical construction and resolves the same reference the publish job pushed, so it is folded identically. It never surfaced in practice only because it chains off `Container Multi-Arch` through `workflow_run`, and that job fails first.

## Deliberately unchanged

`SIGNER_IDENTITY_RE` (image-signing.yml:24) also interpolates `${{ github.repository }}`, but it is the OIDC signer identity rather than an image reference. It must match the repository exactly as GitHub reports it in the token claim, so folding it would break signature verification. A regression test now pins that distinction so the two uses do not get "consistently" fixed later.

## Gating tests

`tests/unit/deploy/test_container_multiarch.py` and `tests/unit/deploy/test_image_signing.py` already assert workflow contents, so they are the natural home for this. Neither asserted anything about `IMAGE_NAME`, which is how the bug survived. Three tests added:

- each workflow contains the fold, and it appears **before** the first reference that consumes it
- the OIDC identity stays unfolded

I verified the new gates can actually fail: with the workflow change reverted and the tests kept, both fold assertions fail; with it applied, all pass.

## Verification

- `pytest tests/unit/deploy tests/unit/release` → 216 passed.
- Full suite → 12,511 passed, 114 skipped, 0 failed.
- `pre-commit run --files` on all four → all hooks pass, including `check yaml` on both workflows.
- Fold behaviour checked against bash 3.2, the oldest shell available, to confirm it carries no bash 4 dependency:

```
PouyanJay/openmed          -> ghcr.io/pouyanjay/openmed
maziyarpanahi/openmed      -> ghcr.io/maziyarpanahi/openmed
Mixed-Case/OpenMed         -> ghcr.io/mixed-case/openmed
```

The upstream value is byte-identical, so this is a no-op for this repository's own builds and only unbreaks forks.

## Data provenance

No data. Workflow configuration and test assertions only.

## Related

Third of three fixes for scheduled and push workflows that fail on `master`: #2915 (nightly `Model Release Gates`) and #2916 (nightly `Fuzz`). All three are independent and touch disjoint files.


maziyarpanahi:
Pouyan, thank you for the careful diagnosis here, especially for keeping the OIDC signer identity untouched while normalizing only GHCR image references. I added the maintainer regression-test alignment on your branch so the image-attestation contract now checks the actual lowercase subject output. The exact head passes the 17 focused deploy and release tests, scoped hooks, the full hosted matrix, and a clean merge-tree rehearsal against current master. I am squash-merging this now.

PR #2946, opened 2026-08-22
perf: speed up MLX privacy-filter inference

anudit:
## Changes

- **Cached Viterbi transition tables** (`openmed/core/decoding/viterbi.py`): $O(C^2)$ BIOES start/end/transition matrix built once per label space instead of every decode call.
- **Vectorized Viterbi DP**: Optional numpy fast path (`scores[:, None] + transitions`) replacing the pure-Python triple loop (~3.2M iterations/step); bit-exact vs the reference across 40 randomized trials, pure-Python fallback retained.
- **Local span-window grapheme snapping** (`openmed/core/decoding/spans.py`): `trim_span_whitespace` walks clusters inside `[start, end)` only — no more full-document cluster rescan per entity span (~59 full-text scans per decode before).
- **Memoized break classes**: `_grapheme_break_class` / `_is_indic_consonant` LRU-cached per code point; `_in_ranges` plain loop instead of generator.
- **Removed redundant GPU→CPU transfer** (`openmed/mlx/inference.py`): Dropped full-matrix `probs = exp(log_probs)` and its `tolist()`; scores recovered as `exp(log_prob)` per predicted token.
- **Opt-in kernel fusion**: `mx.compile` behind `compile_forward=True` / `OPENMED_MLX_COMPILE=1` (default off; ULP-level drift possible).

## Verification

- Bit-exact parity suites for both the DP and span rewrites.
- **218** `mlx`/`viterbi`/`span` unit tests pass.
- No change to CPU PyTorch path or outputs.

---

## Metal (MLX) Latency — Before vs. After

| Batch × Seq Len | Before | After | Improvement |
| :--- | :--- | :--- | :--- |
| **1 × 32** | 135.3 ms | 30.7 ms | 4.4× |
| **2 × 32** | 258.5 ms | 49.1 ms | 5.3× |
| **4 × 32** | 503.4 ms | 80.4 ms | 6.3× |
| **1 × 64** | 280.2 ms | 45.4 ms | 6.2× |
| **2 × 64** | 549.6 ms | 77.8 ms | 7.1× |
| **4 × 64** | 1087.9 ms | 151.3 ms | 7.2× |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -168,6 +168,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   token-throughput, install-size, and peak-RSS records, plus fail-closed
   footprint budgets and archived ARM64 proxy evidence.
 
+### Changed
+
+- Optimized MLX Privacy Filter decoding with bounded BIOES transition-table
+  reuse, equivalent NumPy and pure-Python paths, span-local grapheme work,
+  binary32-compatible confidence reconstruction, and opt-in kernel compilation
+  controls (#2946).
+
 ## [2.2.0] - 2026-08-21
 
 OpenMed 2.2 completes the trustworthy clinical-data-exchange milestone across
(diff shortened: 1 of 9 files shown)


maziyarpanahi:
Thanks @anudit for the contributions! 🙏

PR #3072, opened 2026-09-08
feat(multimodal): add pre-decode limit profiles for multimodal assets

DrVelvetFog:
# Pull Request

## Description
Adds one pre-decode admission policy for multimodal assets, evaluated over the privacy-safe asset manifest before any decoder or model is loaded. `LimitProfile` carries inclusive ceilings for bytes, pages, pixels per image/page/frame, total pixels, frames, and audio duration; `MOBILE_V1` and `DESKTOP_V1` are the agreed initial defaults; `evaluate_asset_limits()` returns deterministic `LimitFinding` records with an allowlisted field name and reason code, the ceiling, and the observed number or null. Unevaluable rules are reported as `insufficient_metadata`, never assumed safe. Follows the model agreed in #2956 point by point.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature which changes existing behavior)
- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test addition/improvement

## Changes Made
- `openmed/multimodal/asset_limits.py`: `LimitProfile` (immutable, re-validated on `with_limits()`), `MOBILE_V1` / `DESKTOP_V1`, `LimitFinding`, `evaluate_asset_limits()`, checked pixel arithmetic bounded by `MAX_PIXEL_PRODUCT`.
- Option (a) for PDF: `PDF_V1` untouched; bytes and pages are checked; the two pixel rules are always reported as `insufficient_metadata` with `observed: null`, and raster pixels are never inferred from page count or from geometry the manifest must not carry. Inapplicable rules (for example audio pages) produce no finding.
- Rules by modality: bytes on all four; pages for PDF; per-unit and total pixels for image (`w × h`) and DICOM (`w × h`, `w × h × frames`); frames for DICOM; duration for audio. Findings come back in a documented, deterministic order.
- The desktop page and pixel ceilings match `render_pdf.py`'s defaults; that provenance and the unbenchmarked status of the other values are documented. Renderer behaviour is unchanged.
- The finding record is separate from `ValidationFinding` and from the abstention reason codes. Exports added to `openmed.multimodal`; docs page, nav and publication entries; CHANGELOG.
[...]

Issue #2956, opened 2026-08-25
Add pre-decode limit profiles for multimodal assets


maziyarpanahi:
## Summary

Large page counts, pixel dimensions, frame counts, or audio durations can exhaust memory before edge inference begins. Existing limits are tied to individual renderers; multimodal callers need one pre-decode policy that evaluates a common asset manifest.

## Scope

- Define immutable mobile and desktop limit profiles for bytes, pages, pixels per page, total pixels, frames, and audio duration.
- Evaluate a privacy-safe asset manifest into stable violation codes, supporting caller-supplied profiles while keeping conservative local defaults.

## Acceptance criteria

- Boundary tests cover values below, equal to, and above every limit for image, PDF, DICOM, and audio manifests.
- Multiple violations are returned in deterministic order and include only field names, limits, and observed numeric metadata.
- Missing optional dimensions are reported as `insufficient_metadata` when a rule cannot be evaluated; the checker never assumes the asset is safe.
- `.venv/bin/python -m pytest tests/unit/multimodal/test_asset_limits.py -q` passes.

## Out of scope

- Allocating tensors, decoding media, benchmarking devices, or auto-downsampling inputs.
- Claiming the defaults fit every model or hardware target.

## Files

- `openmed/multimodal/asset_limits.py`
- `tests/unit/multimodal/test_asset_limits.py`
- `docs/multimodal/asset-limits.md`

---
Task: OM-007e  ·  Milestone: v3.2  ·  Priority: P1  ·  Size: M
Depends on: OM-005e  ·  Blocks: —
Roadmap: `PLANS/V3/OPENMED_V3_HORIZON_BRIEF.md` sections 2 and 4; bounded on-device memory and latency.

DrVelvetFog:
Hi @maziyarpanahi — I'd like to take this one, and then #2980 on top of it. The preflight report is blocked only on this issue now; its other five dependencies (OM-005e/006e/015e/016e/017e) have all landed.

Two things I hit reading the committed tree that I'd rather settle before writing the schema:

[...]

PR #3073, opened 2026-09-08
feat(multimodal): add privacy-safe multimodal preflight report

### LINKED ISSUE #2980: Build a privacy-safe multimodal preflight report
## Summary

The new asset manifest, media-type detector, modality profiles, resource limits, and streaming digest checks need one pre-decode entry point. Without a common report, each future VLM, OCR, DICOM, waveform, and audio provider will assemble different safety behavior.

## Scope

- Orchestrate manifest validation, declared-versus-detected media type, modality-specific fields, digest matching, and resource limits into a single accept-or-abstain report.
- Return deterministic findings, reason codes, schema versions, numeric metadata, and digests only; never open a heavy decoder or copy source content into the report.

## Acceptance criteria

- Synthetic happy-path examples for image, PDF, DICOM, and audio are accepted with byte-stable JSON.
- Media mismatch, malformed manifest, digest mismatch, insufficient metadata, and every resource-limit class produce ordered fail-closed findings.
- A sentinel test proves filenames, paths, OCR text, transcripts, DICOM values, credentials, and raw prefix bytes do not appear in reports or exceptions.
- `.venv/bin/python -m pytest tests/unit/multimodal/test_preflight.py -q` passes.

## Out of scope

- Decoding media, running a model, downsampling inputs, or choosing clinical thresholds.
- A compliance certification or guarantee that accepted media is benign.

## Files

- `openmed/multimodal/preflight.py`
- `tests/unit/multimodal/test_preflight.py`
- `docs/multimodal/preflight.md`

---
Task: OM-018e  ·  Milestone: v3.2  ·  Priority: P1  ·  Size: M
Depends on: OM-005e, OM-006e, OM-007e, OM-015e, OM-016e, OM-017e  ·  Blocks: —
Roadmap: `PLANS/V3/OPENMED_V3_HORIZON_BRIEF.md` sections 2, 4, and 6; shared local-first multimodal safety.

DrVelvetFog:
Closes #2980. Builds on #3072 (OM-007e); the first commit here is that PR's commit and this one applies on top of it. Once #3072 lands I'll rebase so only the preflight commit remains.

## What this adds

`openmed/multimodal/preflight.py` is the single pre-decode entry point the issue asks for. `preflight_asset(manifest, source, *, limit_profile=DESKTOP_V1)` runs the committed contracts in a fixed order and returns one `PreflightReport`:

| Order | Check | Contract | Fail-closed findings |
| --- | --- | --- | --- |
| 1 | `manifest` | `AssetManifest` | `malformed_manifest` — preflight stops, the source is never touched, nothing submitted is reflected |
| 2 | `media_type` | `detect_media_type` | `mismatch`, `unknown` (an undetectable prefix is never a match) |
| 3 | `metadata` | `validate_manifest_metadata` | the profile's own reason codes, plus `unsupported_modality` for `application/dicom+json` |
| 4 | `limits` | `evaluate_asset_limits` | `limit_exceeded`, `insufficient_metadata` |
| 5 | `digest` | `digest_asset` | `byte_count_mismatch`, `sha256_mismatch`, `not_evaluated` |

- The report accepts only when every check passed. Anything else abstains with a `preflight`-stage `AbstentionRecord`: `resource_limit` when the earliest failing check is `limits`, `unsupported_media` otherwise. The detailed findings carry the precise cause.
- `PreflightFinding` has the `check / reason_code / field_name / limit / observed` shape. Metadata and limit findings are validated by constructing the underlying `ValidationFinding` / `LimitFinding`, so no reason code can be invented at this layer.
- The digest pass reads at most one byte past the declared size and runs only when the byte-size ceiling was evaluated and passed, so preflight never reads more than the limit profile's byte ceiling. Seekable streams are restored, non-seekable streams are hashed through a replay of the already-read prefix, caller streams are never closed.
- `to_json()` is byte-stable (fixed key order, no whitespace). Source failures raise `PreflightError` with a stable category and no `__cause__`/`__context__` attached, following `digest.py`'s pattern.

## Per your ruling on #2956

The PDF happy path abstains. A PDF whose bytes, pages, media type, and digest all pass still returns `resource_limit` with two `insufficient_metadata` findings for `pixels` and `total_pixels`. That is tested with an exact JSON string and documented as the intended result rather than weakened into an accept.

One mapping you may want to override: a digest mismatch and a profile violation both abstain as `unsupported_media`, because the preflight stage only allows `unsupported_media`, `resource_limit`, and `provider_unavailable`. I did not touch the abstention allowlist. If you'd rather add a preflight reason for "declared and observed disagree", that's a small follow-up on `abstention.py`.

## Tests

`tests/unit/multimodal/test_preflight.py`: 64 cases. Happy paths for image, DICOM, and audio with exact JSON; the PDF abstain with exact JSON; malformed manifest (six shapes) with proof the source is not read; media mismatch and three undetectable prefixes; profile findings ordered ahead of the limit findings they cause; every resource-limit class; the byte-size ceiling skipping the digest with only the 132-byte prefix read; short and long sources; seekable restore and non-seekable replay; three categorical source failures with no exception chaining; multi-check ordering and the coarse-reason rule; finding and report allowlists (16 rejected shapes); and the sentinel test over paths, OCR text, a transcript, a DICOM patient name, a credential, and raw prefix bytes across reports, reprs, findings, and exceptions.

Locally: 825 passed across `tests/unit/multimodal` plus the docs publication and pages-manifest contracts; ruff check and format clean.

## Docs

`docs/multimodal/preflight.md`, nav and publication entries, CHANGELOG. The staged Pages artifact was measured with `scripts/docs/stage_pages.py`; `budgets.json` moves the file count to 466 and the total/unique ceilings by the new page's footprint with the usual 110,750-byte headroom. The JSON ceiling raised in #3072 still holds (search index at 3,157,823 of 3,407,872).


maziyarpanahi:
Thank you @DrVelvetFog. I reviewed this against #2980 and current `master`, then bounded prefix reads and mutable-buffer handling, required consistent acceptance evidence, and closed the media-type text field. Your contributor commits are preserved.

Validation: full `pytest tests/ -q` — 15864 passed, 133 skipped, 24 warnings in 411.23s (0:06:51); canonical Ruff formatting/lint, scoped mypy, pre-commit security checks, and the complete documentation build passed. The combined release candidate passed 16,547 tests with 133 skips, plus package, repository-policy, license, and documentation checks.

The reviewed head is `8c56124b3a3ddb8477f52b62e0ff36d6e9459c19`. Hosted platform, Nix, security, and documentation checks passed. The inherited wheel-size budget failure also occurs on `master`; the reviewed shared baseline correction in `db7ef5d8dc002d73969c39ab029a60f2f814a0c1` resolves it. The final wheel passes size, installed-profile, import, and PEP 561 checks in Linux. This is ready for v2.5 and will use a normal merge under the five-commit limit.

PR #3143, opened 2026-09-14
fix(clinical): reject duplicate cues in status vocabularies

### LINKED ISSUE #3104: Reject duplicate cues in clinical status vocabularies
## Summary

A duplicated normalized cue can make status matching order-dependent and hide configuration mistakes from contributors.

## Scope

- [ ] Extend status-vocabulary validation to compare cues after the existing Unicode and whitespace normalization.
- [ ] Raise a concise error naming the vocabulary and conflicting status values without echoing clinical text.

## Acceptance criteria

- [ ] Exact, case-only, Unicode-equivalent, and repeated-whitespace duplicates are rejected.
- [ ] The same normalized cue may not silently appear under two statuses.
- [ ] Existing status-vocabulary tests plus focused new tests pass offline.

## Out of scope

- Changing cue priority, adding clinical terminology, or altering normalization results.

## Files

- [ ] `openmed/clinical/status_vocab.py`
- [ ] `tests/unit/clinical/test_status_vocab.py`

---
Roadmap: `PLANS/V3/OPENMED_V3_HORIZON_BRIEF.md` sections 8 and 9.

---
Release plan: #3233
Dependencies: maintained with GitHub issue relationships.
Architecture: Extend `openmed/clinical`, `openmed/eval`, `openmed/mlx`, `openmed/service`, and `openmed/mcp`; generative output stays generative-last behind local backends and release gates.
Estimate: M (planning only; confirm during implementation).
Source of truth: this issue body, its parent release epic, and GitHub dependency relationships.

Mihailchik:
## Description

Status-vocabulary validation now rejects cues that collide after normalization. `_normalize_status` matches through `priority`, so a cue listed under two statuses always resolves to whichever status comes first, and the second listing silently does nothing.

## Type of Change

## Changes Made
- `_validate_vocabulary` passes each status's cues to a new `_reject_duplicate_cues`, which compares them using the existing `_normalize_phrase`: NFKC, hyphen code points folded to `-`, casefold, collapsed whitespace. No second normalization was added, so validation agrees with matching by construction.
- A collision **across statuses** raises `substance vocabulary lists one normalized cue under two statuses: former (cue 1) and current (cue 1)`.
- A collision **within one status** raises `substance.current repeats a cue after normalization (cues 1 and 3)`.
- Errors name the vocabulary, statuses and 1-based cue positions, never the cue text, so the entry can be found without clinical text in logs.
- A cue that is blank after normalization is skipped rather than treated as a collision. It already never matches, and changing that is outside this issue.
- The bundled `status_vocab.yaml` has no collisions — I checked all three vocabularies (48, 35 and 40 cues) before adding the check — so the shipped table still loads unchanged. Normalization results and cue priority are untouched.
- `CHANGELOG.md` entry under Unreleased → Added.

## Testing

New tests in `tests/unit/clinical/test_status_vocab.py`, each loading a synthetic YAML from `tmp_path` through the public `load_status_vocab(path)`:

- across statuses, parametrized: exact, case-only, U+2010 hyphen vs `-`, NFKC full-width, repeated spaces, and a tab;
- within one status, with the positions checked;
- the error text does not contain the cue;
- `part-time` and `part time` stay two distinct cues, as a guard against over-normalizing;
- a synthetic vocabulary without duplicates loads, and so does the bundled one.

On unmodified `master`, 8 of the new tests fail: all six parametrized cases, the within-status one and the no-echo one.

`pytest tests/unit/clinical/test_status_vocab.py`: 44 passed (33 before). `pytest tests/unit/clinical`: 1927 passed, 5 skipped.

## Code Quality
- [x] `ruff check` and `ruff format --check` at the pinned `ruff==0.15.22` pass on both touched files. `status_vocab.py` is not in the scoped mypy file list, so mypy does not cover it.

## Dependencies

## Related Issues
Closes #3104


maziyarpanahi:
Thank you @Mihailchik. This is now on `master` through [#3296](https://github.com/maziyarpanahi/openmed/pull/3296), and #3104 is closed. The batch preserves the duplicate-validation behavior and adds source-free collision diagnostics. Canonical formatting/lint checks passed, along with 229 focused tests and the full suite (17,430 passed, 170 skipped). The final squash tree was audited against the validated batch tree.

PR #3144, opened 2026-09-14
fix(clinical): lint duplicate and conflicting senses in abbreviation inventories

### LINKED ISSUE #3105: Lint duplicate senses in clinical abbreviation inventories
## Summary

Local abbreviation inventories should identify repeated or contradictory sense definitions before they affect deterministic ranking.

## Scope

- [ ] Validate normalized short form, long form, semantic type, and source tuples during inventory loading.
- [ ] Return stable errors for exact duplicates and conflicting definitions.

## Acceptance criteria

- [ ] Case, Unicode, and whitespace-equivalent duplicates are covered by synthetic fixtures.
- [ ] Distinct semantic types remain valid alternatives for the same short form.
- [ ] Focused abbreviation inventory tests pass offline.

## Out of scope

- Expanding the bundled sense inventory or changing the disambiguation scoring algorithm.

## Files

- [ ] `openmed/clinical/abbreviation.py`
- [ ] `tests/unit/clinical/test_abbreviation.py`

---
Roadmap: `PLANS/V3/OPENMED_V3_HORIZON_BRIEF.md` sections 8 and 9.

---
Release plan: #3233
Dependencies: maintained with GitHub issue relationships.
Architecture: Extend `openmed/clinical`, `openmed/eval`, `openmed/mlx`, `openmed/service`, and `openmed/mcp`; generative output stays generative-last behind local backends and release gates.
Estimate: M (planning only; confirm during implementation).
Source of truth: this issue body, its parent release epic, and GitHub dependency relationships.

Mihailchik:
## Description

Abbreviation inventory loading now rejects duplicate and conflicting sense definitions. It also stops losing definitions silently in three cases I reproduced on `master` before changing anything:

| Input | Before | After |
|---|---|---|
| `"CP"`: `chest pain` as `sign_or_symptom` and as `finding` | **1** candidate survives — the merge keyed on long form, so the second overwrote the first | both kept |
| keys `"BP"` (blood pressure) and `"bp"` (bipolar) | **blood pressure silently dropped** — both normalize to `BP`, the later key wins | `ValueError` |
| `"MS"`: `multiple sclerosis` and `Multiple  Sclerosis`, same type | both accepted | `ValueError` |

The first row is the issue's "distinct semantic types remain valid alternatives" criterion. It did not hold on `master`, so it is fixed here rather than only asserted.

## Type of Change

## Changes Made

All in `openmed/clinical/abbreviation.py`:

- **Duplicates and conflicts per short form.** `_reject_duplicate_senses` identifies a candidate by `(_normalize_text(long_form), semantic_type)`. `semantic_type` is already normalized by `_normalize_label` when the definition is parsed, so no new normalization was added.
  - Same identity, same normalized source → `'MS' repeats candidate 1 as candidate 2`.
  - Same identity, different source → `'MS' has conflicting candidates 1 and 3: same long form and semantic type, different source`.
  - Different semantic type → a valid alternative, accepted.
- **Short-form collisions.** `_inventory_from_payload` rejects two raw keys that normalize together: `'BP' and 'bp' both normalize to 'BP'`.
- **Repeated JSON keys.** Both `json.load` calls now pass an `object_pairs_hook` that raises on a key repeated within one object. Without it, `{"BP": [...], "BP": [...]}` keeps the last value before any validation can see it.
- **Merge.** `_merge_inventory` still lets a local definition replace a starter definition with the same long form — `test_user_inventory_overrides_and_extends_starter` passes unchanged. But candidates from one file no longer replace each other: the first same-long-form candidate replaces the existing one, and later ones are appended.
- **The bundled starter inventory has no collisions** — 4 short forms, 12 definitions, checked before adding the validation — so it loads unchanged. Scoring and the bundled data are untouched.
- `CHANGELOG.md` entry under Unreleased → Added.

Errors name the short form and 1-based candidate positions, matching the existing messages in this module, which already quote the short form.

## Testing

New tests in `tests/unit/clinical/test_abbreviation.py`, all with synthetic inventories written to `tmp_path`:

- equivalent duplicates, parametrized: exact, case, whitespace, U+2010 hyphen, NFKC full-width with a source differing only in case;
- a conflict — same long form and type, different source — with positions checked;
- one long form under two semantic types stays two alternatives, loaded without the starter;
- the same, merged over the starter override, so the documented replacement still happens and the second type is appended;
- short forms that normalize together, parametrized: case, surrounding whitespace, full-width;
- a repeated JSON key;
- the starter inventory still loads.

On unmodified `master`, 12 of the new tests fail. I also removed each of the three parts on its own, and each was caught by its own tests:

- without the short-form check → 3 failures;
- with the old merge → 2;
- without the JSON hook → 1.

`pytest tests/unit/clinical/test_abbreviation.py`: 33 passed (20 before). `pytest tests/unit/clinical`: 1929 passed, 5 skipped.

## Code Quality
- [x] `ruff check` and `ruff format --check` at the pinned `ruff==0.15.22` pass on both touched files. `abbreviation.py` is not in the scoped mypy file list.

## Dependencies

## Related Issues
Closes #3105
Related to #3104 — same kind of check for status vocabularies, in #3143


maziyarpanahi:
Thank you @Mihailchik. This is now on `master` through [#3296](https://github.com/maziyarpanahi/openmed/pull/3296), and #3105 is closed. The batch preserves the duplicate-validation behavior and adds source-free collision diagnostics. Canonical formatting/lint checks passed, along with 229 focused tests and the full suite (17,430 passed, 170 skipped). The final squash tree was audited against the validated batch tree.

PR #3251, opened 2026-09-16
fix: isolate tokenizer cache entries by loader

Bembaby:
# Pull Request

## Description
Different loaders requesting the same model/configuration could receive the first loader's tokenizer. Scope cache entries by loader identity while preserving bound-method reuse, LRU eviction, and refresh behavior.

## Type of Change

## Testing
Synthetic baseline: 12 intended failures. Fixed branch: `uv run --frozen --extra dev pytest tests/unit/processing/test_tokenizer_cache_loader_identity.py -q` — 21 passed. Normal `import openmed` passed; Ruff, scoped mypy, and diff checks passed. Full suite reported 29 Windows symlink failures (`WinError 1314`) and 269 skips in tests outside this diff; the full suite was not rerun on unmodified upstream, so these are not claimed pre-existing. No issue is linked until verified.


maziyarpanahi (approved):

PR #3252, opened 2026-09-16
fix: time awaited coroutine execution in profile

Bembaby:
# Pull Request

## Description
The profile decorator timed coroutine construction rather than awaited execution and hid coroutine identity. Keep coroutine functions async and measure awaited execution, including failure and cancellation, without changing the synchronous path.

## Type of Change

## Testing
Synthetic baseline: 7 intended failures. Fixed branch: `uv run --frozen --extra dev pytest tests/unit/utils/test_profile_coroutines.py -q` — 10 passed. Normal `import openmed` passed; Ruff, scoped mypy, and diff checks passed. Full suite reported 29 Windows symlink failures (`WinError 1314`) and 269 skips in tests outside this diff; the full suite was not rerun on unmodified upstream, so these are not claimed pre-existing. No issue is linked until verified.


maziyarpanahi (approved):

PR #3253, opened 2026-09-16
fix: clean up failed evaluation cache writes

Bembaby:
# Pull Request

## Description
Failed evaluation-cache writes could leave `delete=False` temporary reports behind. Track the temporary path and clean it up best-effort in `finally`, preserving the original error, atomic publication, and any existing report.

## Type of Change

## Testing
Synthetic baseline: 7 intended failures. Fixed branch: `uv run --frozen --extra dev pytest tests/unit/eval/test_cache_atomic_cleanup.py -q` — 12 passed. Normal `import openmed` passed; Ruff, scoped mypy, and diff checks passed. Full suite reported 29 Windows symlink failures (`WinError 1314`) and 269 skips in tests outside this diff; the full suite was not rerun on unmodified upstream, so these are not claimed pre-existing. No issue is linked until verified.


maziyarpanahi (approved):

PR #3254, opened 2026-09-16
fix: close Pulsar clients after setup failures

Bembaby:
# Pull Request

## Description
Pulsar client allocation was not unwound when subscription or adapter setup failed. Close the new client best-effort on unsuccessful setup, preserve the setup error, and leave successful ownership with the caller.

## Type of Change

## Testing
Synthetic baseline: 9 intended failures. Fixed branch: `uv run --frozen --extra dev pytest tests/unit/processing/test_pulsar_client_cleanup.py -q` — 14 passed. Normal `import openmed` passed; Ruff, scoped mypy, and diff checks passed. Full suite: 17089 passed, 269 skipped, 29 Windows symlink failures (`WinError 1314`) in tests outside this diff. The full suite was not rerun on unmodified upstream, so these are not claimed pre-existing. No live broker or issue link is claimed.


maziyarpanahi (approved):

PR #3261, opened 2026-09-16
fix: reject unsupported ODT compression methods at the archive gate

### LINKED ISSUE #3246: [BUG] ODT archive gate accepts entries the ZIP reader can't decompress, so extract_odt raises NotImplementedError
### Bug description

`extract_odt` states its rejection contract in its own docstring:

> Raises:
>     UnsupportedDocumentError: If the ODT archive or ``content.xml`` is
>         missing, encrypted, mislabeled, or malformed.

and `tests/fuzz/test_format_parsers_fuzz.py` states the same invariant for every registered parser:

> declared format-rejection errors are acceptable, while every other exception is reported as a parser crash.

A ZIP with a well-formed central directory, the correct ODF `mimetype`, and `content.xml` present — but where an entry declares a compression method Python's `zipfile` doesn't implement — passes the archive gate and then escapes as a bare builtin `NotImplementedError`. A caller that catches `UnsupportedDocumentError`, which is the documented contract, does not catch it.

This is the same shape as #3242: a tolerant gate and a strict reader disagree about what "supported" means.

`_ensure_supported_archive` (`openmed/multimodal/odt.py:346`) is the function that decides an archive is supported. It iterates `archive.infolist()` and checks exactly one thing — the encryption flag bit — then reads `mimetype`. It never looks at `info.compress_type`. `extract_odt` then catches `(OSError, zipfile.BadZipFile, zlib.error)` at `odt.py:337`, and `NotImplementedError` isn't in that tuple.

### Why the one-line fix is the wrong one

Adding `NotImplementedError` to the except tuple at `odt.py:337` makes the traceback go away. I'd argue against it, for two reasons.

**There are two `archive.read()` sites, and the gate itself is one of them.** Putting the bogus method on `mimetype` escapes from `_ensure_supported_archive` at `odt.py:351`; putting it on `content.xml` escapes from `_read_required` at `odt.py:364`. Both are in the repro below. So a fix scoped to `_read_required` would close one and leave the other, and a fix at line 337 has to sit around the whole `with zipfile.ZipFile(...)` block to catch both.

**`NotImplementedError` is a builtin with other meanings.** Catching it around that whole block converts any future genuine `NotImplementedError` raised inside `_ensure_supported_archive` into "not a valid ZIP archive" — a silent misdiagnosis rather than a crash, which is harder to notice than this bug is.

The check belongs in the gate. `_ensure_supported_archive` already walks `infolist()`, so validating `info.compress_type` is one more condition in a loop that is already running, it fails with the documented error before any read is attempted, and it covers both call sites at once. ODF packages use stored or deflate in practice; if that's the intended allowlist, it also rejects bzip2/lzma entries that `zipfile` *would* decompress happily but that aren't ODF — your call on how tight to draw it.

### Minimal reproduction

Fully synthetic, no PHI. Deterministic: the fixture is 544 bytes, sha256 `13980bd0ef88f0ef9bd34cc3444e20e2a0dcc45e08f49d6db4a11c98ac282ca0` for the `content.xml` variant.

```python
import io, struct, zipfile, tempfile, pathlib

MIMETYPE = b"application/vnd.oasis.opendocument.text"
CONTENT = (
    b'<?xml version="1.0" encoding="UTF-8"?>'
    b'<office:document-content '
    b'xmlns:office="urn:oasis:names:tc:opendocument:xmlns:office:1.0" '
    b'xmlns:text="urn:oasis:names:tc:opendocument:xmlns:text:1.0">'
    b"<office:body><office:text><text:p>synthetic</text:p></office:text>"
    b"</office:body></office:document-content>"
)

def build_odt(target: bytes, method: int = 99) -> bytes:
    """A structurally valid ODT ZIP whose `target` entry declares `method`."""
    buf = io.BytesIO()
    fixed = (1980, 1, 1, 0, 0, 0)  # pinned so the fixture is byte-reproducible
    with zipfile.ZipFile(buf, "w", zipfile.ZIP_STORED) as zf:
        zf.writestr(zipfile.ZipInfo("mimetype", fixed), MIMETYPE)
        zf.writestr(zipfile.ZipInfo("content.xml", fixed), CONTENT)
    raw = bytearray(buf.getvalue())
    eocd = raw.rindex(b"PK\x05\x06")
    pos = struct.unpack_from("<I", raw, eocd + 16)[0]
    for _ in range(struct.unpack_from("<H", raw, eocd + 10)[0]):
        name_len, extra_len, cmt_len = struct.unpack_from("<HHH", raw, pos + 28)
        if bytes(raw[pos + 46 : pos + 46 + name_len]) == target:
            local = struct.unpack_from("<I", raw, pos + 42)[0]
            struct.pack_into("<H", raw, pos + 10, method)     # central directory
            struct.pack_into("<H", raw, local + 8, method)    # local file header
        pos += 46 + name_len + extra_len + cmt_len
    return bytes(raw)

from openmed.multimodal import extract_odt
from openmed.multimodal.exceptions import UnsupportedDocumentError

for entry in (b"mimetype", b"content.xml"):
    with tempfile.TemporaryDirectory() as td:
        p = pathlib.Path(td) / "synthetic.odt"
        p.write_bytes(build_odt(entry))
        try:
            extract_odt(p)
            print(f"{entry.decode():12} no error")
        except UnsupportedDocumentError as exc:
            print(f"{entry.decode():12} OK  UnsupportedDocumentError: {exc}")
        except Exception as exc:
            print(f"{entry.decode():12} BUG {type(exc).__name__}: {exc}")
```

Output on `master` `34ebae253`:

```
mimetype     BUG NotImplementedError: That compression method is not supported
content.xml  BUG NotImplementedError: That compression method is not supported
```

The archive itself is not malformed — `zipfile` opens it, lists both entries, and reads `mimetype` back correctly when `mimetype` isn't the patched entry.

### Expected behavior

Both cases raise `UnsupportedDocumentError`, which is what the docstring promises and what the fuzz harness accepts as a declared format rejection.

### Error output

```
  File "openmed/multimodal/odt.py", line 336, in extract_odt
    content = _read_required(archive, _CONTENT_PATH)
  File "openmed/multimodal/odt.py", line 364, in _read_required
    return archive.read(path)
  File ".../zipfile/__init__.py", line 1690, in read
    with self.open(name, "r", pwd) as fp:
  File ".../zipfile/__init__.py", line 1799, in open
    return ZipExtFile(zef_file, mode + 'b', zinfo, pwd, True)
  File ".../zipfile/__init__.py", line 979, in __init__
    self._decompressor = _get_decompressor(self._compress_type)
  File ".../zipfile/__init__.py", line 873, in _get_decompressor
    _check_compression(compress_type)
  File ".../zipfile/__init__.py", line 851, in _check_compression
    raise NotImplementedError("That compression method is not supported")
NotImplementedError: That compression method is not supported
```

### OpenMed version

2.5.0 (reproduced on master `34ebae253` in a clean worktree)

### Operating system

macOS 26.6.2 (arm64); also reproduces on ubuntu-24.04 in CI

### Python version

3.14.6 (CI hit it on 3.11.16)

### Model ID

N/A

### PyTorch version

N/A

### Transformers version

N/A

### Additional context

Your own nightly Fuzz has been reporting `document:odt:.odt` under this test since mid-August — seven runs, all `test_registered_format_parsers_resist_mutated_input`:

| run | date | exception |
|---|---|---|
| [31771561987](https://github.com/maziyarpanahi/openmed/actions/runs/31771561987) | 08-14 | `zlib.error: Error -3 while decompressing data: invalid block type` |
| [31862902448](https://github.com/maziyarpanahi/openmed/actions/runs/31862902448) | 08-15 | same |
| [31925458788](https://github.com/maziyarpanahi/openmed/actions/runs/31925458788) | 08-16 | same |
| [31992909083](https://github.com/maziyarpanahi/openmed/actions/runs/31992909083) | 08-17 | same |
| [32213922964](https://github.com/maziyarpanahi/openmed/actions/runs/32213922964) | 08-19 | same |
| [32445247980](https://github.com/maziyarpanahi/openmed/actions/runs/32445247980) | 08-21 | same |
| [33850974038](https://github.com/maziyarpanahi/openmed/actions/runs/33850974038) | 09-04 | `NotImplementedError: That compression method is not supported` |

`5813082c4` ("fix: reject unreadable ODT entries", 08-22) closed the `zlib.error` path by widening the except tuple, and shipped in v2.5.0. The 09-04 run is after that fix, at the post-fix line numbers (`odt.py:336/364`, `_odt_handler` at 482), and is the one reproduced above — same two `archive.read()` calls, a sibling exception the tuple doesn't name.

Honest scope, so this isn't read as bigger than it is:

- **Narrow trigger.** It needs an archive whose compression-method field is something `zipfile` won't decompress. Real ODT files from LibreOffice/Word don't do this; you get here from a corrupted file, a hostile one, or a producer using an exotic method.
- **Fails closed, not open.** Unlike #3242 there's no text-corruption path — the extraction raises and nothing is silently rewritten. The cost is an unhandled exception type crossing a documented API boundary, so callers doing `except UnsupportedDocumentError` around untrusted uploads get a crash instead of a rejection.
- **Filed public, not as an advisory,** for the same reason as #3242: availability, no PHI disclosure.

If you want the gate-side version, tell me which allowlist you'd draw — stored/deflate, or whatever `zipfile` can decompress — and I'll put the PR up: the check in `_ensure_supported_archive`, plus regression coverage for both entries pinned to the fixture above.

### Privacy confirmation

- [x] I confirm that this issue contains no PHI, PII, secrets, or real patient data.

aycibatuhan:
## Description
`extract_odt` let a bare `NotImplementedError` escape when a ZIP entry declared
a compression method `zipfile` does not implement. The archive gate now rejects
such entries with the documented `UnsupportedDocumentError` before any entry is
read.

## Type of Change

## Changes Made
- `_ensure_supported_archive` already walks `infolist()` for the encryption
  bit. It now also requires every entry's `compress_type` to be stored or
  deflate, which is what ODF packages use.
- The check sits in the gate as the issue suggests, not in an `except` clause.
  That covers both read sites (`mimetype` inside the gate and `content.xml` in
  `_read_required`). `NotImplementedError` is still not caught around the
  archive block, so a real one from our own code will not be misreported as a
  bad ZIP.
- bzip2 and LZMA entries, which `zipfile` can decompress but ODF packages do
  not use, are refused as well. The allowlist mirrors
  `_ALLOWED_DICTIONARY_COMPRESSION` in `openmed/processing/tokenization.py`.

## Testing
- [ ] I have tested this change with different models/inputs

```bash
uv run --frozen --extra dev pytest tests/unit/multimodal/test_odt_extract.py tests/fuzz/test_format_parsers_fuzz.py -q
```

New tests patch method 99 onto `mimetype` and onto `content.xml` in both the
central directory and local header (the repro from the issue), write real
bzip2 and LZMA entries, and confirm a stored `mimetype` next to a deflated
`content.xml` still extracts. The four rejection tests fail on `master` and
pass with this change. The full unit suite passes.

## Documentation
- [ ] I have added docstrings to new functions/classes

The documented contract is unchanged; this makes the code match it.

## Code Quality

## Dependencies

## Checklist
- [x] I have read the contributing guide, Code of Conduct, and maintainer guide

## Related Issues
Closes #3246


maziyarpanahi (approved):
Thank you for this thoughtful, well-scoped contribution. I re-reviewed the repaired exact head and verified the complete hosted check matrix is green. The implementation, tests, documentation, and safety boundaries are all in good shape. Approved for merge.

PR #3262, opened 2026-09-16
fix: treat any address-header parse failure as invalid in redact_email

### LINKED ISSUE #3247: [BUG] Address-header validity probe propagates AttributeError, so redact_email crashes on a nested group header
### Bug description

`_address_header_is_valid` (`openmed/multimodal/email.py:706`) is a probe. Its docstring says so:

> Return whether the stdlib accepts an address without parser defects.

It answers that by assigning the header onto a throwaway `Message` inside a `try`, and catching `(HeaderParseError, IndexError, KeyError, TypeError, ValueError)` at `email.py:713`. On Python 3.11 an address header using nested group syntax makes the stdlib raise `AttributeError`, which isn't in that tuple, so the probe raises instead of returning a bool. It escapes `_redact_headers` (`:749`) and `redact_email` (`:431`) and reaches the caller as `AttributeError: 'Group' object has no attribute 'local_part'`.

The parser fuzz module states the contract for every registered parser:

> declared format-rejection errors are acceptable, while every other exception is reported as a parser crash.

`AttributeError` is not among `_EXPECTED_REJECTIONS`, which is why your nightly reports it as a crash rather than a rejection.

The minimal trigger is an address header value of `a:b:;;`. Under the hood, `email._header_value_parser.get_address_list("a:b:;;").addresses[0].all_mailboxes` yields a `Group` on 3.11, and `headerregistry.py:347` then reads `mb.local_part` off it.

**This is not a report that `f6a6241a1` failed.** That commit ("fix: handle malformed email address headers", 08-22) introduced this probe together with a regression test pinning `b"From: Synthetic Clinic <clinic@"`, and it fixed what it aimed at — your 08-22 03:53Z nightly had failed on exactly that input with `HeaderParseError`, and it doesn't fail on it now. The nested-group value takes a different path through the stdlib. The contrast is in the repro below.

### Why the one-line fix is the wrong one

Adding `AttributeError` to the tuple at `email.py:713` makes this input pass, and I'd argue against it.

The probe is enumerating the private failure modes of `email.headerregistry`. Which exception types that code raises for which malformed input isn't part of its documented API — `AttributeError` here is a stdlib bug surfacing, not a contract — so the tuple can only ever be a list of the cases someone has already hit. It's at five entries after one round of this; `AttributeError` would make six, and the next malformed-header class that reaches a different stdlib code path makes seven.

Since the probe's entire job is to convert "does the stdlib accept this?" into a boolean, I think it should be exception-agnostic: anything raised while parsing an untrusted header value means "not valid," and returns `False`.

The usual objection to a broad `except` is that it hides your own bugs, and I'd normally make that objection myself. It's weak here specifically: the guarded body is two statements, both of them stdlib calls on a throwaway `Message`, with no OpenMed code underneath. The blast radius is the stdlib parse, which is precisely what's being probed.

### Minimal reproduction

Fully synthetic, no PHI. Model-free — `models=identity` mirrors the fuzz harness's `_identity_model`, so nothing is downloaded and the `[hf]` extra isn't needed.

```python
import pathlib, tempfile

from openmed.multimodal import redact_email
from openmed.multimodal.email import _address_header_is_valid

# Mirrors the fuzz harness's _identity_model, so this stays offline and model-free.
identity = lambda text, **_: text

CASES = {
    "truncated  (pinned by f6a6241a1)": "Synthetic Clinic <clinic@",
    "nested group                    ": "a:b:;;",
}

for label, value in CASES.items():
    try:
        result = f"returned {_address_header_is_valid('From', value)}"
    except Exception as exc:
        result = f"{type(exc).__name__}: {exc}"
    print(f"guard   {label} -> {result}")

for label, value in CASES.items():
    eml = f"From: {value}\r\nTo: x@y.invalid\r\nSubject: s\r\n\r\nbody\r\n".encode()
    with tempfile.TemporaryDirectory() as td:
        try:
            redact_email(eml, pathlib.Path(td) / "out.eml", models=identity, lang="en")
            result = "ok"
        except Exception as exc:
            result = f"{type(exc).__name__}: {exc}"
    print(f"public  {label} -> {result}")
```

On `master` `34ebae253`, Python 3.11.14:

```
guard   truncated  (pinned by f6a6241a1) -> returned False
guard   nested group                     -> AttributeError: 'Group' object has no attribute 'local_part'
public  truncated  (pinned by f6a6241a1) -> ok
public  nested group                     -> AttributeError: 'Group' object has no attribute 'local_part'
```

Same file on Python 3.14.6, where the stdlib yields `Mailbox`/`InvalidMailbox` instead of `Group`:

```
guard   truncated  (pinned by f6a6241a1) -> returned False
guard   nested group                     -> returned False
public  truncated  (pinned by f6a6241a1) -> ok
public  nested group                     -> ok
```

Other values that trigger it on 3.11: `g:a:;;`, `a:b:c:;;;`, `A:B:;;`, `g:mb@e.com,h:;;`, `a:b@c,d:;;`. Values that do not: `undisclosed-recipients:;`, `g:;`, `g:a@b;`, `:;` — those parse with defects and the probe returns cleanly.

### Expected behavior

`_address_header_is_valid` returns `False` for `a:b:;;`, and `redact_email` handles it on the same path as any other invalid address header — which is what happens on 3.14 today.

### Error output

```
  File "openmed/multimodal/email.py", line 431, in redact_email
    header_redactions = sum(
  File "openmed/multimodal/email.py", line 432, in <genexpr>
    _redact_headers(part, processor) for part in _iter_mime_parts(message)
  File "openmed/multimodal/email.py", line 749, in _redact_headers
    if normalized in _ADDRESS_HEADERS and not _address_header_is_valid(
  File "openmed/multimodal/email.py", line 711, in _address_header_is_valid
    candidate[header_name] = value
  File ".../email/message.py", line 437, in __setitem__
    self._headers.append(self.policy.header_store_parse(name, val))
  File ".../email/policy.py", line 148, in header_store_parse
    return (name, self.header_factory(name, value))
  File ".../email/headerregistry.py", line 604, in __call__
    return self[name](name, value)
  File ".../email/headerregistry.py", line 192, in __new__
    cls.parse(value, kwds)
  File ".../email/headerregistry.py", line 346, in parse
    [Address(mb.display_name or '',
  File ".../email/headerregistry.py", line 347, in <listcomp>
    mb.local_part or '',
AttributeError: 'Group' object has no attribute 'local_part'
```

### OpenMed version

2.5.0 (reproduced on master `34ebae253` in a clean worktree)

### Operating system

macOS 26.6.2 (arm64); also reproduces on ubuntu-24.04 in CI

### Python version

3.11.14 locally; CI hit it on 3.11.16. Does not reproduce on 3.14.6.

### Model ID

N/A

### PyTorch version

N/A

### Transformers version

N/A

### Additional context

Your nightly Fuzz has reported `document:eml:.eml` three times, and the split lines up with the 08-22 fix:

| run | date | exception |
|---|---|---|
| [32550216681](https://github.com/maziyarpanahi/openmed/actions/runs/32550216681) | 08-22 03:53Z | `HeaderParseError: expected ':' at end of group display name but found '<clinic@'` |
| [33184393259](https://github.com/maziyarpanahi/openmed/actions/runs/33184393259) | 08-28 | `AttributeError: 'Group' object has no attribute 'local_part'` |
| [34327194424](https://github.com/maziyarpanahi/openmed/actions/runs/34327194424) | 09-09 | same |

`f6a6241a1` landed 08-22 07:34Z, between the first row and the second.

On the version boundary, stating only what I measured: it reproduces on 3.11.14 and not on 3.14.6. `headerregistry.py` is byte-identical between those two, so the behaviour change is in `email._header_value_parser` — `all_mailboxes` stops yielding `Group`. **I did not test 3.10, 3.12 or 3.13**, which matters because `requires-python` is `>=3.10` and your CI matrix is 3.10/3.11/3.12 with 3.11 as the default job. There's a related CPython tracker item ([gh-83281](https://github.com/python/cpython/issues/83281) / bpo-39100, "do not treat a Group as invalid-mailbox"); I read it as still open and I haven't confirmed which release changed the behaviour, so I'd treat that link as context rather than as the answer.

That gap doesn't change the recommendation, which is the reason I'd argue for the exception-agnostic version rather than a sixth entry in the tuple: a probe asking "does the stdlib accept this?" shouldn't depend on which stdlib versions raise which type.

Honest scope:

- **Narrow trigger, but reachable.** It needs nested group syntax in a `to`/`from`/`cc`/`bcc`/`reply-to`/`sender` header. No normal mail client emits that; you get here from a corrupted or hostile message — which is the input class `redact_email` exists to handle.
- **Fails closed, not open.** The redaction raises; nothing is written half-redacted and no text is silently rewritten. The cost is that a caller wrapping `redact_email` in the documented rejection types gets a crash instead of a rejection.
- **Filed public, not as an advisory,** same reasoning as #3242 and #3246: availability, no PHI disclosure.

If you want the exception-agnostic version, say so and I'll put the PR up: the probe plus regression coverage pinning `a:b:;;` and the 3.14-vs-3.11 split, so it can't silently start passing for the wrong reason on a newer interpreter.

### Privacy confirmation

- [x] I confirm that this issue contains no PHI, PII, secrets, or real patient data.

aycibatuhan:
## Description
`redact_email` crashed with `AttributeError: 'Group' object has no attribute
'local_part'` on a nested group address header such as `From: a:b:;;`. The
address probe `_address_header_is_valid` now returns `False` for any exception
raised while the standard library parses the value, so the header falls back
to `redacted-address@openmed.invalid` like other malformed addresses.

## Type of Change

## Changes Made
- The probe caught a fixed tuple of exception types. As the issue explains,
  which type `email.headerregistry` raises for a malformed value is not part
  of its API. Adding `AttributeError` would only cover this one input, so the
  probe now uses `except Exception` and returns `False`.
- The guarded body is still two standard-library calls on a throwaway
  `Message`, with no OpenMed code underneath, so the broad `except` cannot hide
  an OpenMed bug. A comment and a `noqa: BLE001` marker explain this, following
  `openmed/core/no_phi_telemetry.py`.
- `KeyboardInterrupt` and `SystemExit` are not `Exception` subclasses and still
  propagate.

## Testing
- [ ] I have tested this change with different models/inputs

```bash
uv run --frozen --extra dev pytest tests/unit/multimodal/test_email.py tests/fuzz/test_format_parsers_fuzz.py -q
```

- Unit tests: the probe returns `False` for `a:b:;;` and for the truncated
  value pinned by f6a6241a1, and `True` for a valid address. `redact_email`
  on a nested group header now produces the fallback address instead of
  crashing.
- Fuzz harness: a registered-parser regression next to the existing
  malformed-header case.
- The three nested group tests fail on `master` (Python 3.11.15) and pass with
  this change. The full unit suite passes.

## Documentation
- [ ] I have added docstrings to new functions/classes

No behaviour change for valid headers; the probe's docstring already describes
the intended boolean contract.

## Code Quality

## Dependencies

## Checklist
- [x] I have read the contributing guide, Code of Conduct, and maintainer guide

## Related Issues
Closes #3247


maziyarpanahi (approved):
Thank you for this thoughtful, well-scoped contribution. I re-reviewed the repaired exact head and verified the complete hosted check matrix is green. The implementation, tests, documentation, and safety boundaries are all in good shape. Approved for merge.

